Issue #103, opened 2022-10-31
How to add new schemas?


psociety:
Thank you for creating PDFMe! It's really cool!

How can i add new schemas?
Seems like the only way is downloading the sources and violating them. Is that the official way?

hand-dot:
Do you mean new type schemas?

> Seems like the only way is downloading the sources and violating them. Is that the official way?
If so, the above way is right.

psociety:
> Do you mean new type schemas?

Yes. Ok, will do it that way then!

By the way:
https://github.com/pdfme/pdfme/blob/f8113fb97817352aa22586193f9286790786c1b1/packages/ui/src/components/Designer/index.tsx#L121
timeTavel lol xD

PR #104, opened 2022-11-02
#103 = Typo

tetreum:
(It doesn't fix the lack of extensibility but the typo signaled in that issue).
https://github.com/pdfme/pdfme/issues/103#issuecomment-1299673748


hand-dot:
Hi  @tetreum .
Thank you!

Issue #112, opened 2022-11-27
Add Rotate UI


hand-dot:
https://daybrush.com/moveable/storybook/?path=/story/basic--rotatable

FikretCin:
Hello there, is this topic under development ?

hand-dot:
Hi @FikretCin 
Not yet... sorry

peteward:
This has been merged and will be released in v3.

PR #132, opened 2023-01-19
[WIP] Add Rotate UI #112

hand-dot:
ref: https://github.com/pdfme/pdfme/issues/112


hand-dot:
This Rotate UI is stuck by Rotate problem on multiline text #142 
First Fix  #142 , after that  Reimplement Rotate UI

PR #151, opened 2023-03-27
Ability to change PDF Size

tiberiusnero560:
Hi there,

I am still quite new to open source, so please forgive any obvious flaws and issues if present. The tests passed for me. The functionality is working when testing. So just need to have it pushed live if others find it useful as well. I'd also like to keep getting updates etc instead of using a forked personal verison. 

Our customers use printers such as the Brother QL-800 label printers. Which can print a single label at a time. This means we cannot do a full sheet. Unless I missed something in the library that can allow me to achieve this. I decided to add the functionality. 

Ability to change the PDF size when calling the generate method.

Added new property to the schema for the generateProps. pdfSize. An array of 2 numbers. 
  pdfSize: z.tuple([z.number(), z.number()]).optional(),
  
  These are used to create the pdf page, and for the pageheight.
      const page = pdfDoc.addPage(pdfSize || [pageWidth, pageHeight]);
                pageHeight: pdfSize ? pdfSize[1] : pageHeight,

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,4 @@
 dist
 coverage/
 __tests__/assets/pdfs/tmp/*.pdf
+/my-app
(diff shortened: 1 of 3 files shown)


tiberiusnero560:
Well I had it working. but this linking is confusing me. Now I cannot get it to work to test it again.

PR #165, opened 2023-05-09
Feature -  Dynamic Font Sizing

steffancarrington:
- Add the ability to dynamically resize the font for text type fields in the UI
- data-testids added to the code base to allow for Jest & Cypress element targeting
- Jest tests added/updated to reflect the code changes. 
- Some semantic HTML improvements made to form labels and inputs

**Dynamic Font Resizing in Action**

https://github.com/pdfme/pdfme/assets/45564901/e6e4c534-5059-47d8-9f8b-218aeb08f843


hand-dot:
Hi, @steffancarrington !! Coool!!
Thank you for implementing Dynamic Font Sizing.

I'll check the code and behavior this weekend.
Please wait for a while and I will reply to you.

steffancarrington:
Hey @hand-dot, 

We also have another PR coming soon to allow dynamic font sizing in the renderer too :)

hand-dot:
I haven't officially reviewed the code yet, but I would like to hear back from you on the following two points.

1. Why doesn't this Pull Request include modifications to the generator? I would like to merge it at the same time as the generator, could you include the changes to the generator in this Pull Request?

2. The modifications to data-testid and testId are not used within the test code, correct? I can't see a reason to include them in the source code.

steffancarrington:
@hand-dot 

1. No problem, we have code modifications for the generator too, we were going to create a separate PR for that tomorrow, but we can update this PR to include the modifications to the generator too 👍 

2. The modifications to the datatestId and testId were added to make elements easier to target with the Cypress Test setup we have in our repo, which uses the PDFMe logic. (My hope was to start adding further data-testid attributes to other areas of the codebase in the future to make it easier for others to also target elements in their tests 🙏 ). - happy to refactor to suit a more general naming convention if this helps? 

Thanks again for taking the time to take a look 👍

hand-dot:
@steffancarrington 

1.
Nice, I'm waiting.

2.
I understand what you are trying to do.
However, please do not include the modification in this PR, but as a separate Issue and consider it for implementation in the future.

I appreciate your contribution. Thank you.

steffancarrington:
@hand-dot 

No problem, will hope to have the updated PR done today/tomorrow ready for your review. 

Thanks again :)

steffancarrington:
Hey @hand-dot , 

The code base has now been updated, I've also moved the logic for the dynamic font sizing into the common package as it's used by both the generate and UI packages. 👍 

When working on the dynamic font sizing I did encounter an issue with the UI (not related to this work), where applying character spacing to text content makes the text content span wider than the resizable content width

Screenshot 2023-05-05 at 08 08 40

I'll create a separate issue for that, and if I get time in the next week or so will implement a fix for it if you're happy for me to do so.

packages/common/src/helpers/calculateDynamicFontSize.ts
```diff
@@ -0,0 +1,80 @@
...
+  const maxSizePercentage = fontSizeScalingMax ?? DEFAULT_FONT_SIZE_SCALING_MAX;
+  const minFontSize = (minSizePercentage * baseFontSizeInPixels) / 100;
+  const maxFontSize = (maxSizePercentage * baseFontSizeInPixels) / 100;
+
+  const doc = await PDFDocument.create();
+  doc.registerFontkit(fontkit);
+
+  let font;
+
+  if (fontName === 'Helvetica') {
```
hand-dot, line 43: Use [DEFAULT_FONT_NAME](https://github.com/pdfme/pdfme/blob/main/packages/common/src/helper.ts#L14).

Please move to constants.ts and export it.

packages/common/src/helpers/calculateDynamicFontSize.ts
hand-dot, line 1: Using pdf-lib for font calculations doesn't seem like a good approach. This function appears to be doing something inefficient.

Couldn't we do something similar with fontkit?

hand-dot:
@steffancarrington 
I've left comments on what I thought as I looked through the code. I will find some time later to run it in the playground repository.

packages/common/src/helpers/calculateDynamicFontSize.ts
steffancarrington: Hey @hand-dot 👋 

I've updated the PR based on your suggestions 👍 

I have however come a little unstuck with the implementation of the font calculations with fontkit alone, would you happen to have any thoughts on how to accomplish this?  

What inefficiencies are occurring with the current setup?

packages/ui/src/components/Designer/Sidebar/DetailView/TextPropEditor.tsx
```diff
@@ -101,7 +132,7 @@ const TextPropEditor = (
   if (activeSchema.type !== 'text') return <></>;
 
   return (
-    <div style={{ fontSize: '0.7rem' }}>
+    <section className="editor-section editor-section--text" style={{ fontSize: '0.7rem' }}>
```
hand-dot, line 135: Is it necessary to add this class name?

packages/common/src/helpers/calculateTextWidthInMm.ts
```diff
@@ -0,0 +1,22 @@
...
+
+export const calculateTextWidthInMm: CalculateTextWidthInMm = (
+  textContent,
+  textFontSize,
+  textFontFamily,
+  textCharacterSpacing
+) => {
+  const characterSpacingWidth = calculateCharacterSpacing(textContent, textCharacterSpacing);
+  const textContentWidthInPt = textFontFamily.widthOfTextAtSize(textContent, textFontSize) + characterSpacingWidth;
+  const textContentWidthInMm = textContentWidthInPt * 0.352778;
```
hand-dot, line 19: Do not use magic numbers.

packages/common/src/helpers/calculateDynamicFontSize.ts
hand-dot: Hi @steffancarrington !

The specific inefficient part is that calculateDynamicFontSize is using pdf-lib.

This is because calculateDynamicFontSize is a method for calculating the font size from the width of the box. It should be possible to implement this without using pdf-lib.

hand-dot:
Hi @steffancarrington !
Finally, I have found some time to do a functionality check, so I will do that now!
I've also added some comments to the review, so I would appreciate it if you could check them out.

Once I finish the functionality check, I will also comment about that.
Thank you as always!

hand-dot:
@steffancarrington 

---

I have conducted a functional test on [pdfme-playground](https://github.com/pdfme/pdfme-playground).

Thank you so much for implementing this feature.

There are three points that need to be addressed:

1. It doesn't work well even when the value for Font Scaling is set with the existing input.

https://github.com/pdfme/pdfme/assets/24843808/cc104c96-2bfe-4f40-b2fa-acb337ea5ba2


2. Changes are not applied when the height is changed by dragging.

https://github.com/pdfme/pdfme/assets/24843808/a1926462-46d4-4692-97a6-3d3a9e2ac406


3. Behavior when the user explicitly makes a line break
    - For texts that have been line-broken, the font size calculation should be applied to the longest text within.

CleanShot 2023-05-21 at 17 09 37@2x



Please feel free to comment if there is a need for discussion about the corrections.

---

As I was trying out this feature, I had some thoughts about text overflow management, so I'd like to leave them as notes.

Currently, there is no implementation of vertical overflow management for the text box, but

Once this feature is merged, there may be a potential for implementing vertical overflow.

Specifically, I think we can take the following measures against text overflow:

1. None: Do nothing. The overflowing text is displayed as is. (Current implementation)
2. Truncation: Truncate. The overflowing text is not displayed.
3. Adjusting Font Size: Change the font size to fit within the box. (Similar to the Shrink text on overflow feature in Google Slide)

ref: [Change how text fits in placeholders & text boxes](https://support.google.com/docs/answer/10364036?hl=en)

This is a feature separate from this pull request, so I will stop any further discussion here,

Thanks to your implementation, I was able to come up with such an idea.

Later on, I would like to create an issue about this.

Thank you.

hand-dot:
I have a question - could you explain why you chose to use Font Scaling Min(%), Font Scaling Max(%)?

Personally, I feel that it would be easier to set up with Dynamic FontSize Min(pt), Dynamic FontSize Max(pt).
I understand, however, that there are disadvantages, such as the behavior becoming slightly complex when changing the font size or inputting initial values.

I would appreciate your thoughts on this matter.

packages/common/src/helpers/calculateDynamicFontSize.ts
steffancarrington: Hey @hand-dot , 

I'll take another look at the fontkit package to see if I can get the dynamic font size calcualtions implemented using fontkit alone. 👍

peteward:
Hi @hand-dot,

The idea for %-based came from me, for a couple of reasons:

1. If the defaults are 100% then if you do not want to use copyfitting then the feature will not apply no matter what size of font you choose. I thought this was preferrable from a UX perspective over having a checkbox to enable the feature.
2. If you decide to change the main font size then the increase or reduction remain relative to that value. (you might want to change both of course if adjusting the main font size though!)
3. I've seen other platforms do it this way, specifically one we're currently using - although it's UX is not good so maybe this was a bad reason :D

steffancarrington:
Hey @hand-dot, 

I've spent some time addressing your feedback in the comments above (thanks again for reviewing the code). 

I have implemented fixes for everything (including multiline support) apart from:

**Using Fontkit alone to calculate the dynamic font size of text content.** 
(I ran into a few issues with implementing this and kept getting errors in our app about the Fontkit module not being found, we currently use CRA & Webpack in our setup, and believe this is a WebPack-related issue) 
The error we're getting is as follows:
> `Attempted import error: 'create' is not exported from 'fontkit' (imported as 'fontkit').`

This occurs when attempting to use the fontkit's `.create()` API as part of the font width calculation logic inside the `@common` package. Have you come across this issue before? 

I did create a CodePen with an example of the font calculation using Fontkit in this [CodePen](https://codepen.io/scarringtonpb/pen/qBJLKRx?editors=1011)


**The issue with Font Scaling Max value not being applied on pre-existing Text elements:** We will be looking to address this after some further testing as it's proven difficult to replicate the issue locally within our App. So I will look at setting up the basePDF set up locally and continue debugging there to come up with a fix for this. (This is a work in progress) 


Once the font scaling issue is fixed, would you be open to merging the PR with the plan of creating a separate issue to re-factor the dynamic font sizing logic to make use of Fonkit alone in the future? 

Internally, for now, we are going to use this implementation as-is (with pdf-lib) because it is blocking us from moving forward.

hand-dot:
Hi @peteward 
Thank you for the reply!
When setting a rule not to go below/above a specified font size,
I thought it was hard to imagine what actual font size you're setting when it's in %.

So, I'm sorry to say this after you've implemented it, but could you change the UI as follows?

1. When not using Dynamic Font Sizing
CleanShot 2023-05-23 at 22 03 07@2x

2. When using Dynamic Font Sizing
CleanShot 2023-05-23 at 21 55 54@2x

hand-dot:
Hey @steffancarrington !

Thank you for taking the time to address the review.

> we currently use CRA & Webpack in our setup, and believe this is a WebPack-related issue)

In my environment, I tried building in the state before reverting the "Calculate Dynamic Font Size using Fontkit" commit, and it built correctly. (nodejs version is 16,18) Also, the built one worked correctly in https://github.com/pdfme/pdfme-playground which I use for debugging.

Are other developers on your team encountering the same error? It would be helpful if you could create a minimal repository that clones the environment where the error is occurring, if possible.

> We will be looking to address this after some further testing as it's proven difficult to replicate the issue locally within our App.

You can easily confirm that the font size of the existing text has been changed and debugging is easy by debugging in the repository of https://github.com/pdfme/pdfme-playground.

> Once the font scaling issue is fixed, would you be open to merging the PR with the plan of creating a separate issue to re-factor the dynamic font sizing logic to make use of Fonkit alone in the future?

Understood. If the same error occurs with your team's developers and that error can be reproduced, I will merge this PR.

Thank you for devoting a lot of time.

peteward:
Hey @hand-dot , 

I understand. We'll change it as per your designs. Thanks for the feedback!

steffancarrington:
Hey @hand-dot, 

I'm sorry for not getting back to you sooner. 

I have now updated the logic in the PR to calculate the dynamic font sizing based on pts rather than percentages. (I have tested this in the PDFMe Playground and it is working as expected for me). 

https://github.com/pdfme/pdfme/assets/45564901/05a0d86b-7cef-43c0-820e-d355d0605dbf

I've also added these additional improvements:

- The dynamic font sizing now uses PT instead of %
- The resizing is triggered as soon as the checkbox is ticked
- It reverts back to using just the font size if the option is unchecked
- The Min Scale value has a max value to match the font size and auto-updates if the font size increases/decreases
- The same happens with the Max Font Scaling where the min value is set to the current font size and auto-updates with the font.

We have an issue with the `fontkit` package being used directly instead of `pdf-lib/fontkit` within our app though (which we embed PDFMe into and are not PDFMe related). We believe the issue is related to a Webpack bundling issue that we have not been able to fix yet. I was just wondering if you have come across this before.

`Attempted` import error: 'create' is not exported from 'fontkit' (imported as 'fontkit').`

Thanks

hand-dot:
Hi @steffancarrington 
Thank you for updating!

I also really appreciate the excellent improvements.
I'll test it when I have time and plan to merge it.

I think we'll probably be able to release it as a new version next week!!

Have a fantastic weekend!

hand-dot:
I have debugged the new version.

I will merge this pull request! Thank you for the great work!

Although it's not a bug caused by this fix, I have discovered the following bug.
[The appearance of the UI preview does not match the look of the PDF generated by the generator. #174](https://github.com/pdfme/pdfme/issues/174)

But this bug frequently occurs when using the 'Dynamic Font Sizing' feature. After the above bug is fixed, I will release it as pdfme v1.2.0.

I will prioritize the aforementioned bug and address it as soon as possible, so please bear with us!

I am sincerely grateful for your contributions. Thank you!

PR #177, opened 2023-06-02
Dynamic Font Resizing - re-added font resizing into generate.ts file …

steffancarrington:
- re-added font resizing into generate.ts file to trigger font resizing on generate when enabled

### DIFF
--- a/packages/generator/src/generate.ts
+++ b/packages/generator/src/generate.ts
@@ -3,10 +3,13 @@ import * as fontkit from 'fontkit';
 import type {
   Font,
   GenerateProps,
+  Schema,
   SchemaInputs,
   Template,
+  TextSchemaWithData,
 } from '@pdfme/common';
 import {
+  calculateDynamicFontSize,
   getDefaultFont,
   getFallbackFontName,
   checkGenerateProps,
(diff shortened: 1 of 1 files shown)


steffancarrington:
Hey @hand-dot, 

When using the latest version of the `main` branch in our app, we noticed that the logic for Dynamically Resizing font in the generate.ts had been removed (https://github.com/pdfme/pdfme/commit/22db786ab1ffb36f58773c9f6ad464ac76a78409#diff-86dc2428e01c4f7437bc94fbcff3a568ff6ca637755855eb8ff08c3ca32b9318) was this intentional? 

Many Thanks,
Steffan

hand-dot:
Hi @steffancarrington 

The modification below was intentional.
https://github.com/pdfme/pdfme/commit/22db786ab1ffb36f58773c9f6ad464ac76a78409#diff-86dc2428e01c4f7437bc94fbcff3a568ff6ca637755855eb8ff08c3ca32b9318

The previously merged code did not support multiple inputs (multiple PDF pages). Therefore, I made a correction.

As you can see from the code below, the calculateDynamicFontSize is used to change the font size according to the user's input characters.
https://github.com/pdfme/pdfme/blob/2868f01449193aeae3af1c764dd4663bf145102b/packages/generator/src/helper.ts#L286

Therefore, I think this pull request is unnecessary.

steffancarrington:
Hey @hand-dot, 

Ah ok, that makes sense, good catch! Thanks for the clarification 👍 

I'll close this PR now, since it's not required. 

Thanks

Issue #191, opened 2023-06-23
Toggle off dynamic font sizing does not update the designer display


peteward:
### Describe the bug

When you un-check dynamic font sizing the changes are not applied immediately. 

### To Reproduce

* Create a large text box
* Enter a small amount of sample text
* Check dynamic font size
* Increase the maximum font size
* The text should dynamically increase in the designer to the new size
* Uncheck dynamic font size
* The text does not return to it's original size

### Expected behavior

* The text should return to it's originally defined size when unchecked.

### Your Environment

```markdown
- pdfme package(@pdfme/ui):
- pdfme version: 1.2.0
- Operating system: Demo designer
- Node.js version or Browser name & version: Chrome 114.0.5735.106
```

### Your Error Log

```shell
-
```

### Additional context

We will contribute the fix for this.

hand-dot:
This bug fix will be included in the next release.

hand-dot:
Released as 1.2.6!

PR #192, opened 2023-06-26
Bugfix - reset text field font size when dynamic font sizing disabled

steffancarrington:
- Bugfix for dynamic font sizing not being reset when disabled in the UI (relates to the issue raised here: https://github.com/pdfme/pdfme/issues/191)


hand-dot:
Hey @steffancarrington ,
I fixed this bug by this [commit](https://github.com/pdfme/pdfme/commit/6b23877d5356a6b575c55809724253de391a551c) .

This bug is caused by the `useEffect` in `TextSchema.tsx` is not considered a toggle off state.

PR #193, opened 2023-06-30
Bug - logic updated to change min font size to match current font siz…

steffancarrington:
- logic updated to change the min font size to match the current font size if the value is greater

**Before**

https://github.com/pdfme/pdfme/assets/45564901/1ad03dd5-9eca-4a5a-a242-af931f4ed829

**After**

https://github.com/pdfme/pdfme/assets/45564901/29e834fb-c8a4-4b62-a840-ad6eec95927e


hand-dot:
@steffancarrington Thanks! LGTM

PR #196, opened 2023-07-17
 Fix text height rendering for PDF generation

steffancarrington:
- Font position adjustment logic added to the UI to ensure it displays in the same places as the generated PDF
- Custom heightOfFontAtSize() method added to correctly render text in generated PDFs. 

**Before**

https://github.com/pdfme/pdfme/assets/45564901/31bf079c-d667-4c62-afe6-381667e7ccd2

**After**

https://github.com/pdfme/pdfme/assets/45564901/b68da68b-34f0-40e1-88b7-f2a31aec9e81


packages/common/src/helper.ts
```diff
@@ -11,6 +11,7 @@ import {
   GenerateProps as GeneratePropsSchema,
   UIProps as UIPropsSchema,
 } from './schema.js';
+import { Font as FontkitFont } from 'fontkit';
```
hand-dot, line 14: please use import type
```
import type { Font as FontkitFont } from 'fontkit';
```

hand-dot (changes requested):
Just minor fix, other than this, Looks good.

hand-dot:
I'll check this PR behavior on the playground from now.
after check, I'll comment.

hand-dot:
Hey @steffancarrington 
Did you try `npm run build` on play-ground?

I got some errors on the production build.

packages/generator/src/helper.ts
```diff
@@ -294,16 +302,25 @@ const drawInputByTextSchema = async (arg: {
...
+
+    const drawLine = (line: string, lineIndex: number) => {
       const textWidth =
-        pdfFontValue.widthOfTextAtSize(splitedLine, size) +
-        (splitedLine.length - 1) * characterSpacing;
-      page.drawText(splitedLine, {
+        pdfFontValue.widthOfTextAtSize(line, size) + (line.length - 1) * characterSpacing;
+
+      const fontkitFont = fontkit.create(
+        Buffer.from(font[templateSchema.fontName as string].data as ArrayBuffer)
```
hand-dot, line 312: templateSchema.fontName is not string, it can be undefined.
if templateSchema.fontName is undefined, we should use a fallback font.

```
      const fontkitFont = fontkit.create(
        Buffer.from(font[templateSchema.fontName ?? fallbackFontName].data as ArrayBuffer)
      );
```

packages/ui/src/components/Schemas/TextSchema.tsx
```diff
@@ -33,46 +34,84 @@ const TextSchemaUI = (
...
       schema.data && schema.backgroundColor ? schema.backgroundColor : 'rgb(242 244 255 / 75%)',
+    border: 'none',
   };
 
+  const schemaFontSize = dynamicFontSize ?? schema.fontSize ?? DEFAULT_FONT_SIZE;
+  let fontAlignmentValue = 0;
+
+  if (schema.fontName) {
+    const currentFont = fontkit.create(
+      Buffer.from(font[schema.fontName].data as ArrayBuffer)
```
hand-dot, line 61: Buffer is not working in the browser.
please install npm 'buffer' and use it.

hand-dot (changes requested):

steffancarrington:
Hey @hand-dot , 

Thanks for reviewing, will look at addressing this feedback today 👍

hand-dot:
Hey @steffancarrington ,
Thanks!

hand-dot:
@steffancarrington  @jbarton123 
Could you please explain what exactly the point of issue in the `pdfme/pdf-lib` library when calculating the `heightOfFontAtSize() `

hand-dot:
I found another problem.
once click text type schema, the text position slightly shifts down from the initial text position.

https://github.com/pdfme/pdfme/assets/24843808/5412874a-ce3a-4674-a1bd-a56b91c54db1

packages/ui/src/components/Schemas/TextSchema.tsx
```diff
@@ -5,13 +5,15 @@ import {
...
   TextSchema,
-  calculateDynamicFontSize,
+  calculateDynamicFontSize
 } from '@pdfme/common';
 import { SchemaUIProps } from './SchemaUI';
 import { ZOOM } from '../../constants';
 import { FontContext } from '../../contexts';
-
+import * as fontkit from 'fontkit';
+import { Buffer } from 'buffer/';
```
hand-dot, line 16:
```
import { Buffer } from 'buffer/';
```
↓
```
import { Buffer } from 'buffer';
```
then you can remove  // @ts-ignore line:61

packages/ui/src/components/Schemas/TextSchema.tsx
```diff
@@ -33,46 +34,85 @@ const TextSchemaUI = (
...
     whiteSpace: 'pre-line',
     wordBreak: 'break-word',
-    border: 'none',
-    color: schema.fontColor ? schema.fontColor : DEFAULT_FONT_COLOR,
     backgroundColor:
       schema.data && schema.backgroundColor ? schema.backgroundColor : 'rgb(242 244 255 / 75%)',
+    border: 'none',
   };
 
+  const schemaFontSize = dynamicFontSize ?? schema.fontSize ?? DEFAULT_FONT_SIZE;
```
hand-dot, line 56: MEMO: Refactoring

hand-dot: I prefer to extract the plain function and move hepler.ts.
and we can create `style` variable from`fontAlignmentValue` of the above function return value .

packages/generator/src/helper.ts
```diff
@@ -294,16 +304,28 @@ const drawInputByTextSchema = async (arg: {
...
-    const drawLine = (splitedLine: string, splitedLineIndex: number) => {
-      const textWidth =
-        pdfFontValue.widthOfTextAtSize(splitedLine, size) +
-        (splitedLine.length - 1) * characterSpacing;
-      page.drawText(splitedLine, {
+    const splitLines = getSplittedLines(inputLine, isOverEval);
+
+    const drawLine = (line: string, lineIndex: number) => {
+      const textWidth = pdfFontValue.widthOfTextAtSize(line, size) + (line.length - 1) * characterSpacing;
+      const fallbackFont = getDefaultFont();
```
hand-dot, line 311: move to the outside of for loop.
this code performance is bad.
```
      const fallbackFont = getDefaultFont();
      let schemaFontData = fallbackFont[DEFAULT_FONT_NAME].data;

      if (templateSchema.fontName) {
        schemaFontData = font[templateSchema.fontName].data;
      }

      const fontkitFont = fontkit.create(Buffer.from(schemaFontData as ArrayBuffer));
```

steffancarrington:
> 



> @steffancarrington @jbarton123 Could you please explain what exactly the point of issue in the `pdfme/pdf-lib` library when calculating the `heightOfFontAtSize() `

Hey @hand-dot ,

The update that was made to the `heightOfFontAtSize()` in the helper was to account for the descent value * the font scale, as this wasn't being considered in the pdf-lib version of the method. 

The change that was introduced was the following (to ensure that we were always taking into account the descender value of the text):

**Before**
  `height -= Math.abs(descent) || 0;`
    
 **After**
  `height -= Math.abs(descent * scale) || 0;`

packages/ui/src/components/Schemas/TextSchema.tsx
steffancarrington: Hey @hand-dot, 

When making this update, we noticed that just using the standard
`import { Buffer } from 'buffer';` was still making reference to the built-in Buffer made available from the NodeJS API and not the export from the `buffer` package.

**Using import from 'buffer'**
Screenshot 2023-07-24 at 12 43 01

**Using import from 'buffer/'**
Screenshot 2023-07-24 at 12 42 37

I could be wrong, but wanted to check first

hand-dot: @steffancarrington 
Oh... You are right!!

https://github.com/feross/buffer#usage

steffancarrington: Hey @hand-dot are we happy to keep this as it is?

packages/generator/src/helper.ts
steffancarrington: I have update this file to move the above further up the `helper.ts` file. This will be visible in my next commit 👍

packages/generator/__tests__/assets/pdfs/assert/Aone31553QRコード.pdf
hand-dot, line 1: @steffancarrington 
I don't understand why a PDF created with a template that does not include Text schema would be updated.

peteward:
Hi @hand-dot,
We did some visual tests with this updated code.
The vertical position of fonts is much better, although not perfect, especially at smaller font sizes.
I don't think there's anything we can do to improve it further though due to the difference in browser font rendering.
[Testing Artwork Builder pdfme fonts.pdf](https://github.com/pdfme/pdfme/files/12179813/Testing.Artwork.Builder.pdfme.fonts.pdf)

packages/generator/__tests__/assets/pdfs/assert/Aone31553QRコード.pdf
steffancarrington: I will revisit these and ensure only PDF's that include TextSchema are updated 👍

PR #202, opened 2023-07-28
Check whether shiftkey is not held on more events to prevent situations where it can get stuck

### LINKED ISSUE #198: Taking a screenshot on mac with CMD+Shift+4 locks selectable elements as if shift is held down
### Describe the bug

After taking a screenshot of an area of the screen, the elements can no longer be interacted with in the normal way, they behave as if shift is being held down.

### To Reproduce

* Load up the designer view with some variable fields visible
* Press CMD+Shift+4 and drag select an area
* Now try to resize one of the text elements and the aspect ratio will be locked as if shift is held down (also if you click another field element it will select both too)

### Expected behavior

- 

### Your Environment

```markdown
- pdfme version: latest
- Operating system: MacOSX Ventura 13.4
```

### Your Error Log

```shell
-
```

### Additional context

Not the highest priority but something to look at one day :)

peteward:
Fixes #198 

Tested on playground. 

It's not 100% perfect as if the first action you do is click to drag a previously selected element it will still remain shifted, but after that it will be gone. I think this is a reasonable compromise for what is an edge case situation anyway.

### DIFF
--- a/packages/ui/src/components/Designer/Main/index.tsx
+++ b/packages/ui/src/components/Designer/Main/index.tsx
@@ -110,7 +110,7 @@ const Main = (props: Props, ref: Ref<HTMLDivElement>) => {
     if (e.shiftKey) setIsPressShiftKey(true);
   };
   const onKeyup = (e: KeyboardEvent) => {
-    if (e.key === 'Shift') setIsPressShiftKey(false);
+    if (e.key === 'Shift' || !e.shiftKey) setIsPressShiftKey(false);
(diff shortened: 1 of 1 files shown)


hand-dot:
Hi @peteward,
Thank you for submitting PR.
Could you resolve conflicts?

Thanks

peteward:
updated 👍

hand-dot:
@peteward 
This fix will be included in the next release
Thank you for your contribution!

PR #203, opened 2023-07-28
Do not lose focus on a textarea in the designer if clicking within it

### LINKED ISSUE #160: Click behaviour on text fields in Designer UI

Currently, it feels a bit confusing when you click on a text field. I expected to double-click to edit which toggles you back out of edit mode. 

Also, if you click once to enter a text field it puts the cursor at the end of the content. I wanted to change content in the middle of the string so I clicked the mouse in the middle like I would a normal text box and it's now no longer in edit mode. I click again and the cursor is back at the end of the string.

I think it might be less confusing if double-click was used to enter/exit the edit mode on a field. 
This way you could use the mouse to position the cursor on the field.

Maybe it's just me though! :) 

Keep it unchanged if other people think it's ok

### Additional context

_No response_

peteward:
fixes #160 

When you click within a textarea that is in focus you expect the text to remain in focus. Common use cases include:
- using your mouse to position the cursor
- highlighting some text for deletion, or copy & paste

Currently, the behaviour is that if you click on a textarea within the designer that is in focus, it will lose focus:

https://github.com/pdfme/pdfme/assets/7068515/bf5b5d70-c6b7-4376-b3c5-8ca2299915ec

New behaviour doesn't deselect the current textarea if clicking within it:

https://github.com/pdfme/pdfme/assets/7068515/ce7db783-2301-4bf9-ba2f-772c6cc44b47

### DIFF
--- a/packages/ui/src/components/Designer/Main/index.tsx
+++ b/packages/ui/src/components/Designer/Main/index.tsx
@@ -195,6 +195,14 @@ const Main = (props: Props, ref: Ref<HTMLDivElement>) => {
     changeSchemas(flatten(arg));
   };
 
+  const currentlyEditingThisTextSchema = (target: EventTarget | null) => {
+    if (!target) return false;
+    if (target instanceof HTMLTextAreaElement)   {
+      return activeElements.map((ae) => ae.id).includes(target.parentElement?.id || '');
+    }
+    return false;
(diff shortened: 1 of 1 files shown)


hand-dot:
Good catch 👍 
Thank you!!

Issue #201, opened 2023-07-27
Use basic templating / placeholder substitution for text schemas


peteward:
### Is your feature request related to a problem? Please describe.

Currently, pdfme only supports dynamic text in added text boxes. If you want to add 'static text' to a text box you have to ensure it is added to the input data, which may not always be viable and doesn't really make sense from an editor perspective.

A common example that we see is `Hi NAME,` where `NAME` comes from the input but the remaining text is static.

### Describe the solution you'd like

Support for very basic variable substitution within content strings to allow for more variability in the text used.

### Describe alternatives you've considered

Also considered a prefixing and suffixing approach #159 , but this felt less clean.

### Additional context

We have a working implementation for this feature, and will open a PR soon once the font vertical positioning code has been merged and any refactoring complete.

It looks and behaves like this from a UI perspective though:

https://github.com/pdfme/pdfme/assets/7068515/730e820e-0455-484f-8f51-dd79487d361c

hand-dot:
I think the concept is good, but I'd like to discuss the interface. Also, I want to have a thorough discussion to avoid adding various features and making it more complex than expected.

Although it's not directly related to this feature, there are also requests for static fields and readonly fields, so I want to consider these changes while keeping these in mind.

Let's discuss in the following discussion.
https://github.com/pdfme/pdfme/discussions/208

peteward:
Closing as agreed to do by custom schema when available

PR #205, opened 2023-07-28
Placeholder text support for mixing dynamic and static text content

peteward:
implements #201 

We're using this feature in our fork of pdfme because it's critical to our functionality (see video on the issue for how it works)

I'm opening this here for feedback and discussion.


hand-dot:
Please check this https://github.com/pdfme/pdfme/issues/201#issuecomment-1658310654

peteward:
Closing this for now, as per the discussion and look towards a future plugins mechanism

hand-dot:
@peteward thank you. we will make better solution 👍

PR #227, opened 2023-08-13
Dynamic font size test fixes

peteward:
Apologies, I'm not sure how I missed this!

Real font test cases updated after edge case fixes.


hand-dot:
Hey @peteward 
No problem! thank you!

PR #229, opened 2023-08-14
TextSchema UI fixes for dynamic font sizing

### LINKED ISSUE #228: Dynamic font resize being called multiple times on UI change
### Describe the bug

This [commit](https://github.com/pdfme/pdfme/commit/ebec6cec6eb822863b16aa41c8a2ad5f2b8c6462) to remove `onDynamicResize` has led to the dynamic font resizing being called 8 times per character typed in the TextShema in the main designer:

I noticed that typing was very janky so added a debug comment to when a character was typed and it went up in increments of 8:
Screenshot 2023-08-13 at 15 11 52

### To Reproduce

Type in a text schema

### Expected behavior

`calculateDynamicFontSize` should only be called once per key press.

### Your Environment

```markdown
- pdfme package(@pdfme/generator or @pdfme/ui): ui
- pdfme version: latest main on playground
```

### Your Error Log

```shell
-
```

### Additional context

_No response_

peteward:
fixes #228 

The dynamic font resize was being called for every text schema on the page on every key press, the same with the setFontAlignment.

### DIFF
--- a/packages/ui/src/components/Schemas/TextSchema.tsx
+++ b/packages/ui/src/components/Schemas/TextSchema.tsx
@@ -27,14 +27,23 @@ const TextSchemaUI = (
 
   useEffect(() => {
     if (schema.dynamicFontSize && schema.data) {
-      calculateDynamicFontSize({ textSchema: schema, font, input: schema.data }).then(setDynamicFontSize);
+      calculateDynamicFontSize({
+        textSchema: schema,
+        font,
+        input: schema.data,
+        startingFontSize: dynamicFontSize,
+      }).then(setDynamicFontSize);
     } else {
       setDynamicFontSize(undefined);
     }
   }, [
-    schema,
     schema.data,
-    schema.dynamicFontSize,
+    schema.width,
+    schema.height,
+    schema.dynamicFontSize?.min,
+    schema.dynamicFontSize?.max,
+    schema.characterSpacing,
+    schema.lineHeight,
     font
   ]);
 
@@ -44,7 +53,6 @@ const TextSchemaUI = (
       setFontAlignmentValue(fav);
     });
   }, [
-    schema,
     schema.width,
     schema.height,
     schema.fontName,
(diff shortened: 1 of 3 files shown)


packages/ui/src/components/Schemas/TextSchema.tsx
```diff
@@ -27,14 +27,23 @@ const TextSchemaUI = (
...
+        textSchema: schema,
+        font,
+        input: schema.data,
+        startingFontSize: dynamicFontSize,
+      }).then(setDynamicFontSize);
     } else {
       setDynamicFontSize(undefined);
     }
   }, [
-    schema,
```
peteward, line 35: Listening on parent objects will result in this useEffect being triggered even if that schema has not changed.
The same is true of listening on `schema.dyanmicFontSize`. 

Removing these and only listening to the attributes explicitly ensures we only trigger when relevant changes are made

packages/ui/src/components/Schemas/TextSchema.tsx
```diff
@@ -27,14 +27,23 @@ const TextSchemaUI = (
 
   useEffect(() => {
     if (schema.dynamicFontSize && schema.data) {
-      calculateDynamicFontSize({ textSchema: schema, font, input: schema.data }).then(setDynamicFontSize);
+      calculateDynamicFontSize({
+        textSchema: schema,
+        font,
+        input: schema.data,
+        startingFontSize: dynamicFontSize,
```
peteward, line 34: The idea of the previous change was initially to solve an issue of efficiency.

When calculating the dynamic font size it was based on an arbitrary value, which is the `schema.fontSize` before you check the dynamicFontSize checkbox.  This means that the font will be iteratively grown or shrunk from this arbitrary value on every change, which is very inefficient, especially as the strings get long.

My idea was to update the `schema.fontSize` with the dynamically calculated value so that the starting point should usually be very close the the new calculated value, which means less iterations in the calculation code.

This alternative passes in a `startingFontSize` which is what we are currently using as the dynamic font size - so the end result for the UI should be the same - better performance.

hand-dot:
Perfect. Thank you!

PR #230, opened 2023-08-14
Dynamic font sizing - Fit options

peteward:
Implemented roughly as per comments on #222 

Before I tidy up and update tests, @hand-dot please can you confirm that you are happy with the behavior - especially around height:

https://github.com/pdfme/pdfme/assets/7068515/e279afd2-8742-4bcf-89c6-0f00cdc8324e


hand-dot:
@peteward 

This behavior is perfect...!
I've been confused about the default values, between vertical and horizontal. 
What do you think?

It's really cool.

CleanShot 2023-08-14 at 22 47 27@2x


CleanShot 2023-08-14 at 22 47 09@2x

packages/common/src/font.ts
```diff
@@ -232,6 +235,9 @@ export const calculateDynamicFontSize = async ({ textSchema, font, input }: { te
   } else if (dynamicFontSize > dynamicFontSizeSetting.max) {
     dynamicFontSize = dynamicFontSizeSetting.max;
   }
+  if (!('fit' in dynamicFontSizeSetting)) {
+    dynamicFontSizeSetting.fit = DEFAULT_DYNAMIC_FIT;
```
hand-dot, line 239: in this function, if there is no setting, just use the default value and don't assign it. 

User should not want the data passed by the template to be changed by reference assignment.

hand-dot (changes requested):

peteward:
Confused about whether it should default to vertical or horizontal?
In our use case we would more likely use vertical, but to avoid BC we might need to go for horizontal?

hand-dot:
> Confused about whether it should default to vertical or horizontal?
> In our use case we would more likely use vertical, but to avoid BC we might need to go for horizontal?

Originally, we should set the default to horizontal, considering backward compatibility (BC), but I have been thinking that the default might be better as vertical since other users would primarily use vertical as well...

There is no backward compatibility, but since this is a feature that has only been released for about two weeks, and we have provided a 'fit' configuration option, we can still accommodate the horizontal use case. Can we set the default to vertical, please?

I apologize for changing my opinion partway through.

peteward:
no problem 😄

peteward:
updated for review now with tests

packages/common/src/font.ts
```diff
@@ -259,22 +263,47 @@ export const calculateDynamicFontSize = async ({
         boxWidthInPt,
       });
       lines.forEach((line) => {
-        const textWidth = widthOfTextAtSize(line, fontKitFont, size, characterSpacing);
-        const textWidthInMm = pt2mm(textWidth);
+        if (dynamicFontFit == DYNAMIC_FIT_VERTICAL) {
```
hand-dot, line 266: use `===` instead of `==`

packages/common/src/font.ts
```diff
@@ -259,22 +263,47 @@ export const calculateDynamicFontSize = async ({
...
+          // For vertical fit we want to consider the width of text lines where we detect a split
+          const textWidth = widthOfTextAtSize(line, fontKitFont, size, characterSpacing);
+          const textWidthInMm = pt2mm(textWidth);
+          totalWidthInMm = Math.max(totalWidthInMm, textWidthInMm);
+        }
 
-        totalWidthInMm = Math.max(totalWidthInMm, textWidthInMm);
         totalHeightInMm += textHeightInMm;
       });
+      if (dynamicFontFit == DYNAMIC_FIT_HORIZONTAL) {
```
hand-dot, line 275: use `===` instead of `==`

packages/common/src/font.ts
```diff
@@ -259,22 +263,47 @@ export const calculateDynamicFontSize = async ({
...
+    return totalWidthInMm > boxWidth || totalHeightInMm > boxHeight;
+  };
+
   let { totalWidthInMm, totalHeightInMm } = calculateConstraints(dynamicFontSize);
 
-  // Attempt to increase the font size up to when it would exceed the box height
-  // (width is also constrained when splitting lines)
-  while (totalHeightInMm < boxHeight && dynamicFontSize < dynamicFontSizeSetting.max) {
+  // Attempt to increase the font size up to desired fit
+  while (shouldFontGrowToFit(totalWidthInMm, totalHeightInMm)) {
```
hand-dot, line 306: Nice

packages/generator/__tests__/assets/templates/dynamicFontSizeVertical.json
```diff
@@ -58,7 +58,8 @@
         "fontName": "NotoSerifJP-Regular",
         "dynamicFontSize": {
           "min": 13,
-          "max": 36
+          "max": 36,
+          "fit": "vertical"
```
hand-dot, line 62: I want to check that it works with the default value as "vertical" without specifying "fit".

packages/ui/src/components/Designer/Sidebar/DetailView/TextPropEditor.tsx
```diff
@@ -97,13 +100,14 @@ const SelectSet = (props: {
   label: string;
   value: string;
   options: string[];
+  width?: string;
```
hand-dot, line 103: All `NumberInputSet` have `width` prop.
we don't need to be optional.

peteward: This is `SelectSet`, not all of them have a width set (Alignment and FontName do not).
We could change this though if you prefer?

hand-dot: Oh... sorry.
so you don't need to change.

peteward:
updated

hand-dot:
@peteward thank you!

PR #232, opened 2023-08-14
Fix switching selections between an in-focus schema and another schema in the UI.

peteward:
This appears to be due to Selecto somehow inhibiting onClick events under certain conditions (including when a textarea is in focus)

The solution was to move the logic currently in top-level onClick into Selecto.onSelect, as this is definitely called.  

https://github.com/pdfme/pdfme/assets/7068515/c975887f-9453-4c83-ab0c-35d3cb7af23d

### DIFF
--- a/packages/ui/src/components/Designer/Main/index.tsx
+++ b/packages/ui/src/components/Designer/Main/index.tsx
@@ -196,14 +196,6 @@ const Main = (props: Props, ref: Ref<HTMLDivElement>) => {
     changeSchemas(flatten(arg));
   };
 
-  const currentlyEditingThisTextSchema = (target: EventTarget | null) => {
-    if (!target) return false;
-    if (target instanceof HTMLTextAreaElement) {
-      return activeElements.map((ae) => ae.id).includes(target.parentElement?.id || '');
-    }
-    return false;
-  };
-
   const onResize = ({ target, width, height, direction }: OnResize) => {
     if (!target) return;
     const s = target.style;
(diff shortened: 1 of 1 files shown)


packages/ui/src/components/Designer/Main/index.tsx
```diff
@@ -240,20 +232,7 @@ const Main = (props: Props, ref: Ref<HTMLDivElement>) => {
   };
 
   return (
-    <div
-      ref={ref}
-      onClick={(e) => {
-        e.stopPropagation();
```
peteward, line 246: I'm not sure if it's still necessary to keep a `stopPropagation()` here?

hand-dot:
LGTM!

packages/ui/src/components/Designer/Main/index.tsx
hand-dot: I can't remember...
I was able to confirm that it works correctly with the image type schema, so I decided that removing it was not a problem.

hand-dot:
@peteward 
On a PullRequest, if you press ".", the VSCode editor of github.dev comes up,
but do you think codesandbox is necessary? 

https://github.com/pdfme/pdfme/assets/24843808/b76ffc1b-ba9f-4c6b-b31d-071f941f1881


Also I came up with an idea, but what do you think about putting the code of the playground repository into pdfme?
I thought it could be easily tested in [codespaces](https://github.com/features/codespaces).

peteward:
I'm using Webstorm not VSCode. I didn't check the codesandbox links, but I will next time!

I think it would be a good idea to include the playground in pdfme - it makes the project more accessible to other collaborators.

PR #235, opened 2023-08-16
[FIX] [TYPO] porps -> props

aadilghani1:
### What's changed
- Fixed a typo on Paper.tsx component

### Why
- "props" is the right word
 
### How
- Updated porps to props


hand-dot:
Good catch👍👍

PR #238, opened 2023-08-18
[FIX] Use hr instead of div for divider

aadilghani1:
What's changed?
- Added hr instead of div for Divider component

Why?
- [Semantically](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/hr) appropriate

How?
- Replacing the div tag with hr tag and retaining the border color as the default color varies according to different browsers


hand-dot:
Hey @aadilghani1 
Thank you.
Please update jest snapshot.

PR #241, opened 2023-08-18
Add an onBlur function to SchemaUIs to allow an action when focus is lost in edit mode

### LINKED ISSUE #240: Further edge cases when leaving editing a text schema and being unable to reselect initially
### Describe the bug

There have been a few attempts at improving the behavior of clicking and selecting on the UI. 

There is at least one currently edge case remaining where clicking between elements will make you initially unable to focus and edit a text schema in the Designer.

### To Reproduce

* Create a text schema
* Start editing the text within the schema in the designer view
* Click on an element in the sidebar (e.g change font size)
* Now click back on the text schema
* The schema will not immediately be selectable, you need to click off it (e.g. on the background) then back on it again.

### Expected behavior

You should be able to move between and reselect elements easily.

### Your Environment

```markdown
- pdfme version: latest main
- Operating system: MacOSX
- Node.js version or Browser name & version: 18, Chrome
```

### Your Error Log

```shell
-
```

### Additional context

We should consider whether it's possible to handle this on leaving the schema element, rather than trying to catch every other element on the page that you could enter (if possible).

peteward:
fixes #240 

Instead of trying to catch every other element that you might click to cancel editing, use the onBlur method of the element you are leaving to take it out of edit mode. 

This means that it doesn't matter where you click when leaving the focus of the textbox or textarea, it will be properly deselected for the next time you click it.

Questions:
* is `onBlur` a good name for the prop? Would it be better as `onStopEditing()` or similar?
* Is it better to make it an optional prop and check it exists before using it, or make it required and pass an empty function when used in `<Preview>`?

### DIFF
--- a/packages/ui/src/components/Designer/Main/index.tsx
+++ b/packages/ui/src/components/Designer/Main/index.tsx
@@ -327,6 +327,7 @@ const Main = (props: Props, ref: Ref<HTMLDivElement>) => {
             onChange={(value) => {
               changeSchemas([{ key: 'data', value, schemaId: schema.id }]);
             }}
+            onBlur={() => setEditing(false)}
             outline={hoveringSchemaId === schema.id ? '1px solid #18a0fb' : '1px dashed #4af'}
             ref={inputRef}
           />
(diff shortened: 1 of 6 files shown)


hand-dot:
Thank you. I'll make the necessary corrections on my end and then merge the changes. 
We plan to release the version with these updates as 2.1.0 during this holiday.

- I believe "onStopEditing" seems to have more versatility, so I'd like to go with "onStopEditing".
- While we might change it in the future, we'll decide against making it optional. Instead, we'll pass an empty function when using it in the <Preview> component.

hand-dot (approved):

hand-dot:
Released as https://github.com/pdfme/pdfme/releases/tag/2.1.0

peteward:
thanks @hand-dot !

Issue #213, opened 2023-08-03
Custom Schema Plugin


hand-dot:
# TaskList

- [x] #246
  - Enable processing from the registered schema in the generator package's renderer (for internal use only)
  - Enable processing from the registered schema in the ui package's renderer (for internal use only)
      - Designer
      - Form
      - Viewer
  - Delete the Sample data editor in ui package's Designer
  - Enable processing from the registered schema in ui package's Designer's Custom prop editor (for internal use only)
- [x] #245
  - Publicize internal-use-only processing and allow for injection of processing from the outside
      - generator
      - ui
          - Designer
          - Form
          - Viewer
  - Move existing processing other than Text to a separate package (pdfme/schema)
      - Barcode
      - Image
  - Confirm that the pdfme/schema package can be injected from the outside
- [x] #283

## Blocker

- [x] Before allowing everyone to write plugins, we need to resolve the [Rotate issue](https://github.com/pdfme/pdfme/pull/277). #142 #112
  - Currently, Rotate is not supported in Designer. Additionally, there is a bug in Generator where text elements cannot be properly rotated.

---

# Custom Schema 1st Draft

## Purpose

- To be able to process schema types that are not supported, by utilizing Custom Schema.
- The processing of previously supported schema types such as text, image, and various barcodes will be transferred to the newly created @pdfme/schema package.
- Users can develop their own Custom Schemas with reference to @pdfme/schema, allowing them to address various issues.

## Modules

Pdfme provides not only PDF rendering but also multiple modules such as Designer's WYSIWYG, Property Editor, Form, Viewer, etc. Users can achieve the goal by adding Custom Schema processing to the required modules.

<details>
<summary><s>@pdfme/common</s></summary>

  - Custom Validator
      - Enable Custom Schema-compatible validation in zod's validation called in the [ui constructor](https://github.com/pdfme/pdfme/blob/b5c176b4981186e1d91811c04bdda2a350d17487/packages/ui/src/class.ts#L80), [inside ui's updateTemplate](https://github.com/pdfme/pdfme/blob/b5c176b4981186e1d91811c04bdda2a350d17487/packages/ui/src/class.ts#L115), and [preprocessing of generator](https://github.com/pdfme/pdfme/blob/b5c176b4981186e1d91811c04bdda2a350d17487/packages/generator/src/generate.ts#L46).
</details>

<details>
<summary>@pdfme/generator</summary>

- PDF renderer
[...]

PR #242, opened 2023-08-19
Enable processing from the registered schema in the generator package's renderer (for internal use only)

hand-dot:
implement one of  https://github.com/pdfme/pdfme/issues/246
ref: https://github.com/pdfme/pdfme/issues/213

---

What was done in this PR are two things. 

The first was creating something called rendererObj, which allows for the future expansion of PDF creation processes. 
The second was the creation of the `generator/src/renderers` folder to separate the PDF rendering processes for each schema.


packages/generator/src/generate.ts
```diff
@@ -65,24 +45,26 @@ const generate = async (props: GenerateProps) => {
...
-          page,
-          fontSetting,
-          inputImageCache,
-        });
+
+        if (!templateSchema || !input) {
+          continue;
+        }
+
+        // TODO ↓ If a Custom Schema is defined, it can undergo custom rendering processes by merging it with the renderer object.
```
hand-dot, line 56: By extending the rendererObj, users can execute processes they have defined.

hand-dot:
I am aware that this PR conflicts with the following PR, but I would like to merge this PR first if the fix is still time-consuming.
https://github.com/pdfme/pdfme/pull/256

Similar adjustments will need to be made to the UI, and I believe that those changes will be significant, similar to this PR.

hand-dot:
Hey @peteward @steffancarrington ,

The review is optional. I have no intention of taking up your valuable time.

However, since it is a significant revision, I have assigned it to you because I want you to understand what I am trying to do and to let me know if there is anything amiss.

Thank you.

peteward:
> I am aware that this PR conflicts with the following PR, but I would like to merge this PR first if the fix is still time-consuming. #256

Hi @hand-dot, I can work on the fixes to #256 later today / tomorrow.  If you still want to merge this PR first then we will rebase and work it out!

I will read this PR now. 👀

packages/generator/src/generate.ts
```diff
@@ -1,57 +1,37 @@
...
-  embedAndGetFontObj,
-  InputImageCache,
-} from './helper.js';
-import { TOOL_NAME } from './constants.js';
-
-const preprocessing = async (arg: { inputs: SchemaInputs[]; template: Template; font: Font }) => {
-  const { template, font } = arg;
+import type { GenerateProps, Template, } from '@pdfme/common';
+import { checkGenerateProps, } from '@pdfme/common';
+import buildInRenderer from './renderers';
```
peteward, line 5: The correct English here would be `builtInRenderer`.

packages/generator/src/generate.ts
```diff
@@ -65,24 +45,26 @@ const generate = async (props: GenerateProps) => {
...
-          inputImageCache,
-        });
+
+        if (!templateSchema || !input) {
+          continue;
+        }
+
+        // TODO ↓ If a Custom Schema is defined, it can undergo custom rendering processes by merging it with the renderer object.
+        // const renderer = Object.assign(buildInRenderer, options.schema);
+        const rendererObj = buildInRenderer[templateSchema.type];
```
peteward, line 58: is the `builtInRenderer` more like `rendererRegsitry` or possibly `compositeRegistry`?  Is it only built-in renderers or will all of them be in this array?

Also, I can't see `builtInRenderer` actually declared or exported anywhere.

packages/generator/src/generate.ts
```diff
@@ -65,24 +45,26 @@ const generate = async (props: GenerateProps) => {
...
+        if (!rendererObj) {
+          throw new Error(`Renderer for type ${templateSchema.type} not found`);
+        }
+        await rendererObj.renderer({ input, templateSchema, pdfDoc, page, options, _cache });
```
peteward, line 62: would `render()` be a better function name (a verb) than `renderer` which is the a noun for the object type?

packages/generator/src/renderers/index.ts
```diff
@@ -0,0 +1,14 @@
+import type { Renderer } from '../types';
+import textRenderer from './textRenderer';
+import imageRenderer from './imageRenderer';
+import barcodeRenderer from './barcodeRenderers';
+import { BarCodeType } from '@pdfme/common';
+
+const barCodeTypes: BarCodeType[] = ['qrcode', 'japanpost', 'ean13', 'ean8', 'code39', 'code128', 'nw7', 'itf14', 'upca', 'upce', 'gs1datamatrix']
+
+const renderer: Renderer = {
+    text: { renderer: textRenderer, },
```
peteward, line 10: ok, so this is where the different renderers are mapped 👍 

perhaps: 
``` 
   text: { render: renderText }, 
   image: { render: renderImage }.`
```
etc? 

this naming feels more functional rather than classes/objects, wdyt?

packages/generator/src/types.ts
```diff
@@ -0,0 +1,23 @@
...
+import type { GeneratorOptions, Schema, } from '@pdfme/common';
+import type { PDFPage, PDFDocument, } from '@pdfme/pdf-lib';
+
+export type EmbedPdfBox = {
+    mediaBox: { x: number; y: number; width: number; height: number };
+    bleedBox: { x: number; y: number; width: number; height: number };
+    trimBox: { x: number; y: number; width: number; height: number };
+};
+
+export interface RenderProps {
```
peteward, line 11: `RendererProps` ?

peteward (commented):
very cool and exciting! looks very clean and logical to me

packages/generator/src/generate.ts
hand-dot: It is more like rendererRegistry. 
In the future We allow users to add their own rendering processes to this registry using options.

packages/generator/src/renderers/index.ts
hand-dot: You are right!
Since it is just a function rather than a renderer, we will use Render.

Also, I will rename textRenderer.ts and imageRenderer.ts to textRender.ts and imageRender.ts.

Issue #190, opened 2023-06-23
Vertical text alignment


peteward:
### Is your feature request related to a problem? Please describe.

We recently contributed dynamic text sizing, however, the growing or shrinking of text is from a baseline position and you will likely need to control this based on the layout of your page.

e.g.:
![Screenshot 2023-06-23 at 12 27 32](https://github.com/pdfme/pdfme/assets/7068515/ee6c5376-fa54-4e05-9ea7-09c4a83dabd9)

text gets shrunk dynamically, but it "moves up" from the desired position (just above the text below) because there is an implicit top alignment to text boxes:

![Screenshot 2023-06-23 at 12 27 59](https://github.com/pdfme/pdfme/assets/7068515/74ef51a1-d18b-46dd-a047-48db18039ea5)

Another use for this feature is that it is sometimes useful if you want to perfectly center text within an available space, to be able to draw a textbox between 2 horizontal planes then just align the text.
![Screenshot 2023-06-23 at 12 30 44](https://github.com/pdfme/pdfme/assets/7068515/3fcda686-c2f6-4461-be98-51691ffdce2b)

### Describe the solution you'd like

Add a vertical alignment option to all text boxes (even without dynamic sizing set). This would contain the options of top, middle, bottom and would adjust the text to the appropriate position within the box.  It would also be respected by dynamic font sizing.

The default option would be top, so all existing templates and schemas would be unaffected.

### Describe alternatives you've considered

-

### Additional context

We will start looking into this contribution.

peteward:
Something simple like this:

![Screenshot 2023-06-23 at 13 07 13](https://github.com/pdfme/pdfme/assets/7068515/51127177-801b-48b5-81be-8ea4a30a1d06)

Alternative labels:

![Screenshot 2023-06-23 at 13 08 21](https://github.com/pdfme/pdfme/assets/7068515/273c3c65-4d64-400e-86c3-0ef670b2fc2e)

hand-dot:
Please use this label.
![](https://user-images.githubusercontent.com/7068515/248263476-273c3c65-4d64-400e-86c3-0ef670b2fc2e.png)

hand-dot:
ref: ref: https://github.com/Hopding/pdf-lib/issues/645#issuecomment-718029496

jameshalsall:
@hand-dot we are still working on this issue, was there a reason you linked to the above comment?

hand-dot:
Hi @jameshalsall 
[...]

PR #247, opened 2023-08-26
Vertical alignment support for text schema

peteward:
implements #190 

### Example:

https://github.com/pdfme/pdfme/assets/7068515/6cbad2ce-fa7b-4248-9f4d-daa6192bdf48

### Notes & Caveats

Vertical aligning in edit mode in the designer is surprisingly challenging. You cannot vertically align textarea, so we had to dynamically resize it after any change to ensure it's set to the height of it's content and position it within the parent container (credit to @steffancarrington).  This does lead to some minor side-effects - sometimes, with certain fonts, the scrollbar is visible when it should not be. Also, there's occasionally a 1 pixel shift between edit and view mode due to the rounding of `textarea.scrollHeight`.

In implementing this we spotted an issue with dynamic font sizing not properly considering line-height which I'll log separately and look into.


hand-dot (approved):

hand-dot:
Released as [2.2.0](https://github.com/pdfme/pdfme/releases/tag/2.2.0) !!!

Issue #215, opened 2023-08-04
Back button when editing a field


peteward:
### Is your feature request related to a problem? Please describe.

Currently, to return to the list of fields when editing a single one, you have to click somewhere in the main designer window - either another field to select, or just on the pdf background to deselect and show the list

### Describe the solution you'd like

It would improve the UX to have a back button to take you back to the list, it could be added in the top left corner where this red rectangle shows:

![pdfme-back-button](https://github.com/pdfme/pdfme/assets/7068515/192b384d-c660-4002-8532-bdb8bff8c344)

Our designer came up with the following concept as an idea:

![Screenshot 2023-08-04 at 15 48 28](https://github.com/pdfme/pdfme/assets/7068515/9dc1ffca-b396-4485-a94c-621ef957aa17)

### Describe alternatives you've considered

-

PR #251, opened 2023-08-29
Add a back button from detail to list views

peteward:
implements #215 

Let me know if you would prefer a different icon.

https://github.com/pdfme/pdfme/assets/7068515/4e9e6888-857d-451f-ac04-dea3c31a66fe


hand-dot:
Cool!

Issue #254, opened 2023-09-04
Colored QR Codes


peteward:
### Is your feature request related to a problem? Please describe.

We have a requirement for changing the color of QR Codes in our designs

### Describe the solution you'd like

Allow the bars, background and preview text colours of QR codes to be controllable through the UI, and to render in the generated PDF.

### Describe alternatives you've considered

-

### Additional context

The work has already been done for this by @steffancarrington, I'm just opening up an issue to describe the feature.

PR #256, opened 2023-09-06
Custom qr code colours

steffancarrington:
This PR addresses the issues raised: https://github.com/pdfme/pdfme/issues/254

- The ability to customise Barcode Colours has been added. 
- Users will be able to change the colours of the following properties on barcodes: **barcolor**, **backgroundcolor** and **textcolor**

- Barcode is generated in the UI as the user updates the colour, background colour, text colour or input values. 

**UI Control Updates for QR Codes and other Barcodes**

https://github.com/pdfme/pdfme/assets/45564901/1b81daab-5daf-4824-9581-bc122c0c35b5

**Changing QR Code Colours**

https://github.com/pdfme/pdfme/assets/45564901/f23e1953-c2ef-424a-aedb-4121f87c26dd

**Changing Barcodes that support Text (UPCA as an example)**

https://github.com/pdfme/pdfme/assets/45564901/250fdd95-b197-43fe-a6f4-38c832212b8d


packages/ui/src/components/Designer/Sidebar/DetailView/BarCodePropEditor.tsx
```diff
@@ -0,0 +1,121 @@
...
+
+const inputStyle = {
+  width: '90%',
+  color: '#333',
+  background: 'none',
+  borderRadius: 2,
+  border: '1px solid #767676',
+};
+
+const ColorInputSet = (props: {
```
peteward, line 20: Just a note here:

This is very nearly a duplicate of the corresponding property in TextPropEditor. We considered moving this to shared code, but I know that this area will likely be reworked when Custom Schema Plugin support is implemented, so I thought it worth getting your opinion first @hand-dot.

Should we consolidate this to a shared file? Or leave them separate for now?

packages/common/src/helper.ts
```diff
@@ -193,6 +193,7 @@ export const validateBarcodeInput = (type: BarCodeType, input: string) => {
     let res = input.match(regexp);
     if (
       res != null &&
+      input.length <= 52 && // 52 is the max length of a GS1 DataMatrix barcode before bwip-js throws an error
```
hand-dot, line 196: nice!

packages/ui/src/components/Schemas/BarcodeSchema.tsx
```diff
@@ -54,18 +35,44 @@ const ErrorBarcode = () => (
...
-    return null;
-  }
+  useEffect(() => {
+    const canvas = document.createElement('canvas');
+    const barcodeType = type === 'nw7' ? 'rationalizedCodabar' : type;
 
-  return validateBarcodeInput(schema.type as BarCodeType, value) ? (
-    <SampleBarcode schema={schema} />
+    try {
+      bwipjs.toCanvas(canvas, {
```
hand-dot, line 49: Real time preview barcode! i like this.

hand-dot: Wait I guess we can extract this creating barcode image process to common package.

I wrote same code in generator package

packages/ui/src/components/Designer/Sidebar/DetailView/BarCodePropEditor.tsx
peteward: Updated.

I think we should probably do the same for other form elements, but in a different PR

packages/ui/src/components/Schemas/BarcodeSchema.tsx
peteward: I have looked at this code and I couldn't easily created a single function here. 
The existing `createBarCode` function returns a `Promise` to a `Buffer` whereas this new code uses the canvas directly. It's also not doing much beyond creating the parameters for the function call.

So instead I have moved some of the logic that creates these parameters into shared functions that both the UI and generator use.

Let me know what you think.

hand-dot:
@steffancarrington @peteward 
Thank you for the cool barcode option. 
I really appreciate this.

hand-dot:
I gonna publish a new version of pdfme that includes this PR after merging [another PR](https://github.com/pdfme/pdfme/pull/242)

packages/ui/src/components/Schemas/BarcodeSchema.tsx
hand-dot: memo:

```
function bufferToDataURL(blob) {
    return new Promise((resolve, reject) => {
        const reader = new FileReader();
        reader.onloadend = () => resolve(reader.result);
        reader.onerror = reject;
        reader.readAsDataURL(blob);
    });
}

const buffer = new Blob([new Uint8Array([/* binary data */])], { type: 'image/png' });
bufferToDataURL(buffer).then(dataURL => {
    console.log(dataURL);
}).catch(error => {
    console.error('Error:', error);
});
```

hand-dot: I successfully moved createBarCode to common.

ref: https://github.com/pdfme/pdfme/commit/8d4b9f00f2fe3ef8f9d44c703f45bdac039fd241

steffancarrington:
Amazing thanks @hand-dot and @peteward  🙌

Issue #269, opened 2023-09-20
checkFont should be more graceful with edge cases


peteward:
### Is your feature request related to a problem? Please describe.

In our situation we upload fonts for use with the designer. 
It's possible that we get into a situation where a font that we saved with the schema is no longer passed in. This gives us a fatal error.

There are a number of conditions in `checkFont` that seem quite aggressive when handling error scenarios.

### Describe the solution you'd like

We could consider adjusting the input data to work with what's provided, rather than throwing Errors.

This is the current code:

```export const checkFont = (arg: { font: Font; template: Template }) => {
  const {
    font,
    template: { schemas },
  } = arg;
  const fontValues = Object.values(font);
  const fallbackFontNum = fontValues.reduce((acc, cur) => (cur.fallback ? acc + 1 : acc), 0);
  if (fallbackFontNum === 0) {
    throw Error(`fallback flag is not found in font. true fallback flag must be only one.`);
  }
  if (fallbackFontNum > 1) {
    throw Error(
      `${fallbackFontNum} fallback flags found in font. true fallback flag must be only one.`
    );
  }

  const fontNamesInSchemas = getFontNamesInSchemas(schemas);
  const fontNames = Object.keys(font);
  if (fontNamesInSchemas.some((f) => !fontNames.includes(f))) {
    throw Error(
      `${fontNamesInSchemas
        .filter((f) => !fontNames.includes(f))
        .join()} of template.schemas is not found in font.`
    );
  }
};
````

* Instead of throwing an error if no Fallback font is specified, we could set the first font as fallback
* If more than one font is set as fallback we could unset all but the first
* If a font used in one of the schemas is missing, use a fallback

I think the only error condition is if there's no font at all that can be used.

### Describe alternatives you've considered

Leave it as it is!

peteward:
won't implement

PR #271, opened 2023-09-21
Improve checkFont to adjust provided input rather than fatal Error

peteward:
implements #269 

Instead of throwing Errors when the input doesn't work, adjust the schema or font settings so that it the situation can recover.

If no font is, or too many are, set as default then we can adjust this and continue processing (with a warning).
Similarly, once we have a fallback set we can use it when a font is missing rather than being unrecoverable.

### DIFF
--- a/packages/common/__tests__/font.test.ts
+++ b/packages/common/__tests__/font.test.ts
@@ -100,9 +100,8 @@ describe('checkFont test', () => {
 
     const font = getFont();
     checkFont({ template: getTemplate(), font });
-    const expectedFont = getFont();
-    expectedFont.SauceHanSansJP.fallback = true;
-    expect(font).toEqual(expectedFont);
+    expect(font.SauceHanSansJP.fallback).toEqual(true);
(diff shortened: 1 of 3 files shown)


hand-dot:
Hey @peteward 

In instances like this PR, where the [existing documents](https://pdfme.com/docs/guides/custom-fonts) are changing, it would be appreciated if updates to the documents on the website are also included.

I understand the issue you're facing, as described in the situation below on https://github.com/pdfme/pdfme/issues/269,

> It's possible that we get into a situation where a font that we saved with the schema is no longer passed in. This gives us a fatal error.

Should this be handled by pdfme? I guess you should check on your application and pass the correct font, or your application should change the template.

Furthermore, utilizing the order of objects as a fallback could lead to unintended behavior since the order isn't guaranteed.
ref: https://stackoverflow.com/questions/5525795/does-javascript-guarantee-object-property-order/38218582#38218582

Could you reconsider why this PR is necessary and what behavior should be implemented?

peteward:
Hi @hand-dot,

Thanks for the review.

No problem, I will move this logic into our app instead 👍

Issue #142, opened 2023-02-06
Rotate problem on multiline text


Yan-Anq:
### Describe the bug

When I rotate a text with multines (for example 90 deg), I obtain an unique line :
![rotate_wish](https://user-images.githubusercontent.com/124576871/216959325-1f183de5-c95b-4b9a-801b-3e529ca946ed.png)

### To Reproduce

I use 'rotate' parameter :

```js
const template = {
  "schemas": [
    {
      "field1": {
        "type": "text",
        "position": {
          "x": 24.87,
          "y": 91.55
        },
        "width": 250,
        "height": 100,
        "alignment": "left",
        "fontSize": 10,
        "characterSpacing": 0,
        "lineHeight": 1,
        "rotate": 90    // <------ Rotation 90deg
      }
    }
  ],
  "basePdf": BLANK_PDF
};
```

### Expected behavior

Each lines must be separate with a 'lineHeight' parameter.

### Your Environment

```markdown
- pdfme package : @pdfme/generator
- pdfme version: ^1.1.5
- Operating system: Docker Windows Desktop (VueJs 3 + Vite)
```

### Your Error Log

```shell
No error in console.
```

hand-dot:
Reproduce by Sample App: https://stackblitz.com/edit/typescript-efv8uk?file=index.ts

Issue #245, opened 2023-08-21
Custom Schema Plugin Step 2


hand-dot:
## Custom Schema Plugin #213

---

- [x] [Move existing processing to a separate package (pdfme/schema)](https://github.com/pdfme/pdfme/pull/278)
- [x] [Publicize internal-use-only processing and allow for injection of processing from the outside](https://github.com/pdfme/pdfme/pull/281)
- [x] Release pdfme/schema package

---

### Blocker

Before allowing everyone to write plugins, we need to resolve the [Rotate issue](https://github.com/pdfme/pdfme/pull/277). #142 #112

Currently, Rotate is not supported in Designer. Additionally, there is a bug in Generator where text elements cannot be properly rotated.

PR #277, opened 2023-10-09
[WIP] Rotate

hand-dot:
This PR needs to be resolved before upgrading to v3.

---

ref: https://github.com/pdfme/pdfme/issues/112
ref: https://github.com/pdfme/pdfme/issues/142

### DIFF
--- a/packages/ui/src/components/Renderer.tsx
+++ b/packages/ui/src/components/Renderer.tsx
@@ -22,6 +22,7 @@ const Wrapper = ({
       width: schema.width * ZOOM,
       top: schema.position.y * ZOOM,
       left: schema.position.x * ZOOM,
+      transform: `rotate(${schema.rotate ?? 0}deg)`,
       outline,
     }}
   >
--- a/packages/ui/src/components/Renderer.tsx
+++ b/packages/ui/src/components/Renderer.tsx
@@ -23,6 +23,8 @@ const Wrapper = ({
       top: schema.position.y * ZOOM,
       left: schema.position.x * ZOOM,
       transform: `rotate(${schema.rotate ?? 0}deg)`,
+      // If we want to avoid logic in other pdf renderers then we would use this line below:
+      // transformOrigin: 'bottom left',
       outline,
     }}
   >
--- a/packages/common/src/schema.ts
+++ b/packages/common/src/schema.ts
@@ -20,7 +20,7 @@ export const Schema = z.object({
   position: z.object({ x: z.number(), y: z.number() }),
   width: z.number(),
   height: z.number(),
-  rotate: z.number().optional(),
+  rotate: z.number().optional(), // FIXME optionalじゃない方がいいかも
 }).passthrough();
 
 const SchemaForUIAdditionalInfo = z.object({
--- a/packages/common/src/schema.ts
+++ b/packages/common/src/schema.ts
@@ -20,7 +20,7 @@ export const Schema = z.object({
   position: z.object({ x: z.number(), y: z.number() }),
   width: z.number(),
   height: z.number(),
-  rotate: z.number().optional(), // FIXME optionalじゃない方がいいかも
+  rotate: z.number().optional(),
 }).passthrough();
 
 const SchemaForUIAdditionalInfo = z.object({
--- a/packages/generator/__tests__/renderUtils.test.ts
+++ b/packages/generator/__tests__/renderUtils.test.ts
@@ -1,0 +1,28 @@
+import { Schema, mm2pt } from '@pdfme/common';
+import { convertForPdfLayoutProps } from '../src/renderUtils';
+
+// npm run test:single -- "convertForPdfLayoutProps"
+
+// FIXME ここから unit testを作る
+describe('convertForPdfLayoutProps', () => {
+    it.only('should return correct value', () => {
+        const schema: Schema = {
+            type: 'image',
+            width: 100,
+            height: 100,
+            position: { x: 100, y: 100 },
+            rotate: 0
+        };
+        const pageHeight = 1000;
+
+        const { position: { x, y }, height, width, rotate } = convertForPdfLayoutProps({ schema, pageHeight });
+
+        expect(height).toEqual(mm2pt(schema.height));
+        expect(width).toEqual(mm2pt(schema.width));
+
+        expect(x).toEqual(mm2pt(schema.position.x));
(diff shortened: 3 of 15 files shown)


peteward:
this is really exciting!

hand-dot:
@peteward 
I'm stuck with the implementation of `convertForPdfLayoutProps`... If possible, I would appreciate some help.

hand-dot:
ref: https://github.com/Hopding/pdf-lib/issues/524

hand-dot:
This PR conflicts with the following PR:
https://github.com/pdfme/pdfme/pull/278

If the modifications become more extensive than anticipated, please report here before making the changes.

packages/ui/src/components/Renderer.tsx
```diff
@@ -22,6 +22,7 @@ const Wrapper = ({
       width: schema.width * ZOOM,
       top: schema.position.y * ZOOM,
       left: schema.position.x * ZOOM,
+      transform: `rotate(${schema.rotate ?? 0}deg)`,
```
peteward, line 25:
```suggestion
      transform: `rotate(${schema.rotate ?? 0}deg)`,
      transformOrigin: 'bottom left',
```

Hi @hand-dot, 

Sorry for the slow reply.

I have just been looking at this. I think the problem is that the FE and pdflib are rotating from different positions. 
The FE was rotating from the center, but pdflib is rotating from the bottom-left corner.

I have tried setting the rotation origin as suggested here and the positioning appears to be better:


Screenshot 2023-10-13 at 12 55 27

Screenshot 2023-10-13 at 12 55 59


However, there is still a positional alignment issue as you can see where the text is being aligned to the bottom of the box:

Screenshot 2023-10-13 at 12 58 39
Screenshot 2023-10-13 at 12 59 10

I will continue to look at this and update with my findings.

peteward: This will be quite challenging for multi-line text.
Because we draw each line separately, the point of rotation will be different for each line:

Screenshot 2023-10-13 at 15 09 39
Screenshot 2023-10-13 at 15 10 26

@hand-dot do you remember why we draw each line separately, rather than with a single `page.drawText()` call?

peteward: I have tried joining the lines and rendering them together in a single pdflib call, but this doesn't appear to be much better, because the rotation point is around the top line of the text, not the bottom. Here are some examples with a small rotational increment, you can see they are rotating from the bottom corner of the 'P':

Screenshot 2023-10-13 at 17 17 58
Screenshot 2023-10-13 at 17 18 22
Screenshot 2023-10-13 at 17 20 54

I think we would need to calculate the positional offset of each row and apply them, which will be fun...

peteward: Images and bar codes do rotate as they do in the UI though which is good:

Screenshot 2023-10-13 at 17 23 46
Screenshot 2023-10-13 at 17 23 58

It is only text which is going to be difficult.

peteward: I have some ideas of how to calculate these adjustments for text schemas. I will continue trying in the next few days.

hand-dot: Hey @peteward ,
Amazing!!! Thank you. I appreciate your contribution.

hand-dot: > @hand-dot do you remember why we draw each line separately, rather than with a single page.drawText() call?
we are using separate page.drawText() calls for each line in order to support text alignment.

hand-dot:
Merged the large pull request below and conflicts have already occurred.
https://github.com/pdfme/pdfme/pull/278

It might be best to close this PR and recreate it as a separate one.

hand-dot:
This problem is blocker for  #245

peteward:
@hand-dot, I appreciate this PR is now out of date, but I have fixed the rotation - please see the [latest commit](https://github.com/pdfme/pdfme/pull/277/commits/9e22e8b1d14eb19d39a314be43ff710694534202).

The solution is to rotate the point at which pdflib starts drawing from the point at which the UI performs rotation. We then need to translate the difference between these points to move the rendering into the correct position.


However, there is a decision to make and I wanted to check with you first.

We can either:

1). Rotate in the UI from the center of the schema item (same for text, images, barcodes, etc.):

Screenshot 2023-10-18 at 14 37 20

- better UX
- but we would need code to handle point rotation in every schema

2). Rotate in the UI from the bottom-left corner:

Screenshot 2023-10-18 at 14 39 42

- perhaps less good UX
- but the code is only required in the text schema pdf rendering function, because other schemas will already be rotated from the bottom left corner by pdflib


I think option 1 is best because the UI experience is more important and the additional code to handle rotation is not too complex. What do you think?

hand-dot:
@peteward 

Fantastic, thank you so much! I was completely stumped on how to implement the rotation, so this is a huge help. I really appreciate you as always.

I agree with the first option as well. I'm thinking of exporting the rotation-related logic from the schema package so that it's easily accessible for users.

Would it be possible for you to create a new PR from the latest main branch? If that's too much trouble, I would appreciate it if you could add the rotation logic to the image and barcode renderings; I can then use that code as a reference to update the main branch myself.

peteward:
Hi @hand-dot,

Yes, I will try to create a new PR from latest main.  I have thought of a way to do the rotation cleanly so it should not be a problem, but I will also expose the translation function in case others need to use it.

I am on a busy family vacation this week but will try my best to get it done soon.

hand-dot:
Thank you! @peteward ,

I definitely want you to prioritize family time.

I can work on documentation for V3 until the new PR is created, so please work at your own pace.

Supporting rotation in Designer will allow us to make the 'rotate' option in the Schema non-optional from V3 onwards. 
'rotate’ as option is something I've been personally don’t like, so I'm glad.

hand-dot:
Note:
It might be better to leave "rotate" as an optional feature.
 This is because if "rotate" is set in the [defaultSchema of propPanel each schema](https://github.com/pdfme/pdfme/blob/c2fd0a8faec626dbc3e4e1c621f76241647a9638/packages/schemas/src/text/propPanel.ts#L133), it should be rotatable in the designer. 
of course  if not set, it should not be rotatable in the designer.(current behaver)

Implementing "rotate" as a mandatory feature could complicate future support for more complex schemas like tables, increasing the workload. Therefore, I think it's better to make it configurable whether each schema can be rotated or not.

I will implement this separately from the new Rotate PR.

peteward:
superseded by #282

PR #278, opened 2023-10-10
Add schemas package

hand-dot:
- [x] Move existing processing to a separate package (pdfme/schemas)
    - [x] Text
    - [x] Barcode
    - [x] Image
- [x] Move some pdfme/common functions to pdfme/schemas
- [x] Confirm that the pdfme/schema package can be injected from the outside

---

ref: https://github.com/pdfme/pdfme/issues/245


hand-dot:
Hi @peteward,

I apologize for the very large pull request. Essentially, this update accomplishes the following:

- Created a schemas package
- Moved rendering processing from common package to schemas package
- import rendering processes from schemas to generator/ui

Could you please review the changes before merging?

packages/schemas/src/index.ts
hand-dot, line 1: I believe you'll find it easier to understand if you trace the process from this file.

packages/generator/src/builtInRenderer.ts
hand-dot, line 1: In the future, barcodes will be excluded from the built-in features and will instead be loaded as plugins on the user side as needed. (The same applies to the UI package.)

hand-dot: The following objects are used for their respective purposes and locations:
- ui: Used in Viewer, Form, Designer (`packages/ui/src/builtInRenderer.ts`)
- pdf: Used in Generator (`packages/generator/src/builtInRenderer.ts`)
- propPanel: Used in Designer (`packages/ui/src/builtInPropPanel.ts`)

packages/schemas/src/text/uiRender.ts
```diff
@@ -0,0 +1,153 @@
...
+    backgroundColor: getBackgroundColor(mode, value, schema),
+    border: 'none',
+    display: 'flex',
+    flexDirection: 'column',
+    justifyContent: mapVerticalAlignToFlex(schema.verticalAlignment),
+    width: '100%',
+    height: '100%',
+  };
+  Object.assign(container.style, containerStyle);
+  rootElement.innerHTML = '';
```
peteward, line 93: I see these files have changed from JSX, which feels a bit harder to read. Was it not possible to use JSX?

peteward (approved):
Great work - I understand the changes.
Keep moving forwards and any small things can be iterated 👍

packages/schemas/src/text/uiRender.ts
hand-dot: @peteward 
Indeed, JSX is more readable.

However, requiring users to write plugins in React would mean forcing them to unify the UI package and React versions, as well as pre-transpile JSX.

While React might not be an option, handling UI rendering through direct DOM manipulation is a clear responsibility. Overall, this approach is the simplest.

PR #281, opened 2023-10-19
Publicize internal-use-only processing and allow for injection of processing from the outside

hand-dot:
- [x] Publicize internal-use-only processing and allow for injection of processing from the outside
  - [x] generator
  - [x] ui
      - [x] Designer
      - [x] Form
      - [x] Viewer
  - [x] Prop check

---

ref: https://github.com/pdfme/pdfme/issues/245


hand-dot:
Use like this
```js
import { barcodes } from "@pdfme/schemas"

new Designer({
  template,
  domContainer,
  plugins: {
    qrcode: barcodes.qrcode,
    code128: barcodes.code128
  }
});

// ---

generate({
  template, 
  inputs,
  plugins: {
    qrcode: barcodes.qrcode,
    code128: barcodes.code128
  }
});
```

Issue #285, opened 2023-10-23
Editor schema props form layout makes it hard to see selected font


peteward:
### Is your feature request related to a problem? Please describe.

The new form layout makes all form elements the same width, regardless of content. 
This can be a problem for font which is truncated and makes it hard to see which font has been chosen, especially if you are using lots of variations of the same font family:

<img width="113" alt="Screenshot 2023-10-23 at 15 11 28" src="https://github.com/pdfme/pdfme/assets/7068515/383af595-ec0c-46e6-ae0d-eb38d615c6ff">

### Describe the solution you'd like

Resize the form elements to use the space more appropriately based on their content.

### Describe alternatives you've considered

.

hand-dot:
Nooooo.... thank you for reporting.

peteward:
I will open a PR for this. 

Maybe longer-term there are changes that can be made at a theme level.

hand-dot:
Closed for the moment.

PR #287, opened 2023-10-23
Reorganise text render props for wider font name field

peteward:
Improves #285:

Screenshot 2023-10-23 at 16 26 21


hand-dot:
Great!

PR #290, opened 2023-10-26
Set Rotate from propPanel default schema

hand-dot:
Instead of mandating the implementation of 'rotate' for all elements, which would increase the developer's workload, allow for the omission of 'rotate' implementation on a case-by-case basis.

If 'rotate' is set [here](https://github.com/pdfme/pdfme/blob/8e4ccc3c35c6008e5cf5ec885c90799d68a70f7c/packages/schemas/src/text/propPanel.ts#L137), enable rotate in the Designer, and if 'rotate' is not set here, make rotate unavailable.


packages/ui/src/components/Designer/Main/index.tsx
```diff
@@ -242,6 +242,10 @@ const Main = (props: Props, ref: Ref<HTMLDivElement>) => {
     setEditing(true);
   };
 
+  const rotatable = schemasList[pageCursor]
```
hand-dot, line 245: detect rotatable or not

packages/schemas/src/text/propPanel.ts
```diff
@@ -135,6 +135,9 @@ export const propPanel: PropPanel<TextSchema> = {
     position: { x: 0, y: 0 },
     width: 45,
     height: 10,
+    // If the value of "rotate" is set to undefined or not set at all, rotation will be disabled in the UI.
+    // Check this document: https://pdfme.com//docs/custom-schemas#learning-how-to-create-from-pdfmeschemas-code
+    rotate: 0,
```
hand-dot, line 140: set rotate as number

packages/ui/src/components/Designer/Sidebar/index.tsx
```diff
@@ -1,33 +1,18 @@
...
   const i18n = useContext(I18nContext);
-  const font = useContext(FontContext);
-  const fallbackFont = getFallbackFontName(font);
   const [open, setOpen] = useState(true);
 
-  const getActiveSchemas = () => {
-    const ids = activeElements.map((ae) => ae.id);
-    const activeSchema = schemas.find((s) => ids.includes(s.id));
-
-    if (activeSchema?.type === 'text') {
```
hand-dot, line 22: refactoring

peteward:
nice!

Issue #303, opened 2023-10-31
Separating Schema label from code identifier


peteward:
### Is your feature request related to a problem? Please describe.

Currently, in v3 we use the key of the plugins object for displaying to the user:

```
      type: {
        title: 'Type',
        type: 'string',
        widget: 'select',
        props: {
          options: Object.keys(propPanelRegistry).map((label) => ({ label, value: label })),
        },
      },
```

![Screenshot 2023-10-31 at 07 19 17](https://github.com/pdfme/pdfme/assets/7068515/82ffdeea-abee-4948-9ef5-ead0735b2ef7)

### Describe the solution you'd like

I think these labels could be more user-friendly, e.g.:

```
Text
Image
QR Code
Barcode EAN8
Barcode EAN13
````

I think it would also make sense to have a separation between what we call a plugin in the schema JSON versus what we display to the user.  Ideally, we could then allow developers to rename the label used in certain scenarios. e.g. I want to add a new type of text plugin and call it `Text Advanced` and I would like to call the existing text plugin `Text Simple`, regardless of what they are referenced as within the JSON object.

It is my understanding that the key that we currently use comes from here:

```
      designer.current = new Designer({
        domContainer: designerRef.current,
        template: pdfmeTemplate,
        options: { font: fonts },
        plugins: {
          text,
          image,
          qrcode: barcodes.qrcode,
          ean8: barcodes.ean8,
          ean13: barcodes.ean13,
        },
      });
```

The default label should probably be defined within the plugin (and be translatable in future) but it would be ideal if you could also override it at the point of use above.

I may have misunderstood how this works though, I've only just started looking at writing a custom schema 👍 

### Describe alternatives you've considered

-

hand-dot:
I'll check this.
I’d like to be include this to v3.0.0

hand-dot:
I'm currently working on this. It might take 2-3 days as I need to make more extensive modifications to the entire library than I initially thought. Documentation updates will also be needed.

However, I've found a way to simplify not just the labels but also the internal implementation, 
so I plan to include this in the V3 release.

Thank you for making this proposal.

hand-dot:
We can use like this

<img width="765" alt="CleanShot 2023-11-02 at 00 13 29@2x" src="https://github.com/pdfme/pdfme/assets/24843808/a57b1f04-2cff-419c-9433-ce9ab9bfeb0c">

https://pdfme.com/docs/custom-schemas

PR #304, opened 2023-11-01
Separating Schema label from code identifier

hand-dot:
ref: https://github.com/pdfme/pdfme/issues/303

CleanShot 2023-11-01 at 11 29 59@2x


packages/common/__tests__/helper.test.ts
```diff
@@ -219,7 +219,7 @@ describe('checkPlugins test', () => {
       pdf: async () => {},
       ui: async () => {},
       propPanel: {
-        propPanelSchema: {},
+        schema: {},
```
hand-dot, line 222: This fix is not directly related to this PR.

packages/common/src/helper.ts
```diff
@@ -142,14 +142,11 @@ export const checkPlugins = (arg: { plugins: Plugins; template: Template }) => {
   } = arg;
   const allSchemaTypes = uniq(schemas.map((s) => Object.values(s).map((v) => v.type)).flat());
 
-  // text and image are builtin schema types so they are not included in plugins.
-  const exceptBuiltinSchemaTypes = allSchemaTypes.filter((t) => t !== 'text' && t !== 'image');
+  const pluginsSchemaTypes = Object.values(plugins).map((p) => p?.propPanel.defaultSchema.type);
 
-  const pluginsSchemaTypes = Object.keys(plugins);
```
hand-dot, line 148: Stopped using the 'plugins' key.

packages/generator/src/generate.ts
```diff
@@ -62,7 +55,9 @@ const generate = async (props: GenerateProps) => {
           continue;
         }
 
-        const render = rendererRegistry[schema.type];
+        const render = Object.values(plugins).find(
+          (plugin) => plugin.propPanel.defaultSchema.type === schema.type
```
hand-dot, line 59: There may be users who are only using the generator, but accessing `plugin.propPanel.defaultSchema.type` feels a bit awkward. Should it be accessible through `plugin.type` instead?

packages/generator/src/types.ts
```diff
@@ -1,11 +1,3 @@
-import type { Schema, PDFRenderProps } from '@pdfme/common';
```
hand-dot, line 1: We just use plugin.

This was a welcome side effect.
Thank you for making this proposal.

packages/generator/src/generate.ts
hand-dot: I might be overthinking this. If it doesn't feel awkward as is, I'd like to keep it that way.

peteward (commented):
this is looking a lot cleaner :+1:

PR #305, opened 2023-11-01
Fix alignment of sidebar top buttons

peteward:
These were previously not aligned:

Screenshot 2023-11-01 at 16 50 15


hand-dot:
Nice!

PR #306, opened 2023-11-02
Adds automated publishing workflow to NPM

dkastl:
This pull request introduces an automated publishing workflow to NPM for the multiple packages included in this repository.

The motivation behind this PR is to simplify the use of the multiple NPM packages in this repository through NPM. In the release documentation it seemed to me that the publishing process so far was manual, and I thought automation will allow publishing of NPM packages more frequently.

The publishing process is triggered by creating a Git tag and it distinguishs between stable releases and pre-releases by publishing to different NPM channels. 
Additionally, the CI creates releases to the `dev` channel on NPM for every commit to the main branch, so it's easy to stay up-to-date with the latest development.

I tested the workflow successfully until the actual publishing to NPM, which requires to configure a NPM Authentication Token name `NPM_TOKEN` as Github secret.

### DIFF
--- a/.github/workflows/publish-commit.yml
+++ b/.github/workflows/publish-commit.yml
@@ -1,0 +1,42 @@
+# .github/workflows/publish-commit.yml
+name: Publish Commit
+
+on:
+  push:
+    branches:
+      - main
+
+jobs:
+  publish-npm:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+          
+      - uses: actions/setup-node@v3
+        with:
+          node-version: '20'
+          registry-url: 'https://registry.npmjs.org'
+
+      - name: Increment version and append git hash
+        run: ./scripts/increment-version.sh
+
+      - name: Install dependencies
+        run: npm ci
+
+      - name: Set NPM version
+        run: npm version --no-git-tag-version $VERSION
+
+      - name: Build
+        run: npm run build
+
+      - name: Test
+        run: npm run test
+        env:
+          CI: true
+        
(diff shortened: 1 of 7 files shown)


hand-dot:
@peteward 
I'd like to know your opinion

hand-dot:
I used to update manually, but it seems good that with this script, development releases can be obtained automatically from the main branch.

I would like to adopt this operation as it seems better.

peteward:
sorry for the late reply.

Releases are your area @hand-dot, so if you are happy with the process then great.

Thank you very much @dkastl for the contribution! 👍

hand-dot:
@dkastl 

I'm currently trying to release V3 using this workflow, but it fails with an error. The cause of the error seems to be that the new value is not set correctly with npm version, and it tries to publish with a duplicate version, which prevents the publish from happening.

If you have any ideas for a fix, I'd like to hear them.

https://github.com/pdfme/pdfme/actions/runs/6742936403

hand-dot:
I'm sorry. I didn't understand how to use it.

It seems that 3.0.0-beta.2 already existed, so it was necessary to create the tag from 3.0.1 onwards.

dkastl:
Sorry, I was not connected to internet today, but I will take a look tomorrow. I'm not sure if NPM requires increasing version numbers for example. It's the part I wasn't able to test properly because changing the NPM organisation causes other issues.

dkastl:
It took me a bit longer to take a look, but checking the GH Actions and published NPM packages it seems to working now, right?

Thanks a lot for merging! This makes is a lot easier to stay up-to-date with the latest state of development!

hand-dot:
@dkastl Thank you👍

PR #308, opened 2023-11-03
Ensure schemas are more uniquely cached for generation

### LINKED ISSUE #266: Same colour combination added to multiple QR Codes if URL's are the same
### Describe the bug

This is an edge case which may not happen often but is still worth raising an issue for. 

If a user was to add multiple QR Codes to a PDF with the same URL, but specify different `barcolor` and `backgroundcolor` HEX values, both QR Codes would render out in the same colour, this related to the Preview / Generated PDF as the UI displays both QR Codes correctly.

****

### To Reproduce

1. On an existing PDF, Add a new field with the type of 'qrcode'
2. Select a `barcolor` for the QR Code
3. Select a `backgroundcolor` for the QR Code
4. Enter a URL for the QR Code (http:///www.google.com)
5. Repeat the above steps, but selecting different `barcolor` and `backgroundcolor` values for the second QR Code
6. Keep the URL the same
7. Click the 'Preview PDF' button. 

**Generated PDF with incorrect QR Codes**
Screenshot 2023-09-14 at 12 05 00

### Expected behavior

- Both QR Codes should render in their selected `barcolor` and `backgroundcolor` regardless of the URL specified.

**Correct Rendering of PDF**
Screenshot 2023-09-14 at 12 05 13

### Your Environment

```markdown
- pdfme package(@pdfme/generator or @pdfme/ui): ui
- pdfme version: latest
- Operating system: Mac
```

### Your Error Log

```shell
-
```

### Additional context

No response

peteward:
fixes #266

### DIFF
--- a/packages/schemas/src/renderUtils.ts
+++ b/packages/schemas/src/renderUtils.ts
@@ -61,4 +61,4 @@ export const rotatePoint = (
   return { x, y };
 };
 
-export const getCacheKey = (schema: Schema, input: string) => `${schema.type}${input}`;
+export const getCacheKey = (schema: Schema, input: string) => `${JSON.stringify(schema)}${input}`;
--- a/packages/schemas/__tests__/renderUtils.test.ts
+++ b/packages/schemas/__tests__/renderUtils.test.ts
@@ -2,10 +2,23 @@ import { Schema, mm2pt, pt2mm } from '@pdfme/common';
 import { convertForPdfLayoutProps, rotatePoint, getCacheKey } from '../src/renderUtils';
 
 describe('getCacheKey', () => {
-  it('should return a key of name and input', () => {
+  it('should return a key of JSON string version of the schema with input appended', () => {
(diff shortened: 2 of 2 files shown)


hand-dot:
@peteward 
This change cause performance issue?
Can you check test fail reason?

packages/schemas/src/renderUtils.ts
```diff
@@ -61,4 +61,4 @@ export const rotatePoint = (
   return { x, y };
 };
 
-export const getCacheKey = (schema: Schema, input: string) => `${schema.type}${input}`;
+export const getCacheKey = (schema: Schema, input: string) => `${JSON.stringify(schema)}${input}`;
```
hand-dot, line 64: JSON.stringify should not be used in thread-blocking operations such as within loops.

peteward:
oh sorry, this was stupid. 
I will try again...

PR #311, opened 2023-11-04
Update package.json ?? That dont work without "set" Instruction.

solarpush:
Hi, there,
i try to clone repo for build plugins and customs schema but i can't build the package without the set instruction.

Control if is right way. 

Thanks


solarpush:
I thinks i do false for install the dev environnement

PR #312, opened 2023-11-04
Fix barcode cache key to capture uniqueness properly

peteward:
I went for a simple solution of just editing the barcode code. 

Since we only have 2 types of cacheable schemas right now, doing anything more complicated seemed like over-engineering at this point.

Let me know what you think @hand-dot


peteward:
fixes #266

hand-dot:
LGTM!

Issue #310, opened 2023-11-04
Change build tool from webpack to vite to improve slow pdfme/ui build times


hand-dot:
Perhaps the playground repository should be incorporated into this repository.

PR #314, opened 2023-11-08
Change build tool from webpack to vite to improve slow pdfme/ui build times

hand-dot:
ref: https://github.com/pdfme/pdfme/issues/310


hand-dot:
memo: https://vitejs.dev/guide/build.html#library-mode

hand-dot:
@peteward 

Moved the playground code to this repository.
Also, changed the build system from webpack to vite for both the playground and UI.

By including the playground in the UI, we can make the most of vite's HRM features. However, since it feels odd to include the playground in the UI package, I decided not to do it at this time.

To use it, run npm run dev in the playground dir, and if you're running npm run develop in each package, changes will automatically reflect in the playground. For packages other than UI, the reflection is instant, but for the UI package, it takes about 10 seconds to reflect (on a MacBook Air M2).

I think this has made development easier. If there are no issues with this, I plan to discontinue the playground repository.

peteward:
great job @hand-dot !

PR #316, opened 2023-11-10
Add types to package exports

enberg:
This PR adds types to package exports where "exports" are defined.

This fixes issues when using the packages with moduleResolution set to "bundler" in tsconfig like Next.js defaults to.

More info on how the resolution is affected is available here: https://www.typescriptlang.org/docs/handbook/modules/reference.html#packagejson-exports

To reproduce the error you can:
Create a new Next.js app `npx create-next-app@latest`
Install pdfme common package `npm i @pdfme/common`
Attempt to `import { BLANK_PDF } from '@pdfme/common'`

You should see the following error:

Screenshot 2023-11-10 at 12 41 29

### DIFF
--- a/packages/common/package.json
+++ b/packages/common/package.json
@@ -28,7 +28,8 @@
   "exports": {
     ".": {
       "import": "./dist/esm/src/index.js",
-      "require": "./dist/cjs/src/index.js"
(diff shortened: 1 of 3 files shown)


hand-dot:
@enberg Thank you for contributing!!

PR #318, opened 2023-11-12
add type="button" to all <button> elements

### LINKED ISSUE #317: Many <button> elements submit a form containing the designer

It's the same problem as #280, but concerns all other `<button>` elements (for example the zoom controls currently are not usable for me, since they submit the surrounding form)

The same solution as in #280 needs to be applied to all other <button> elements (`<button [...]` to `<button type="button" [...]`).

I've tried capturing and stopping propagation of the event and forcing the `type="button"` attribute on all `<button>` elements after rendering, but that doesn't work reliably.

### Additional context

_No response_

rhurling:
fixes #317


hand-dot:
@rhurling LGTM! Thank you~

Issue #243, opened 2023-08-21
Custom Design for UI


hand-dot:
TaskList

- [x] Apply ThemeConfig to Viewer, Form, Designer.
- [x] Replace all React component parts under `packages/ui/src/components` with [AntDesign components](https://ant.design/components/overview).
    - [x] Designer/Sidebar/index.tsx
    - [x] Designer/Sidebar/ListView/index.tsx
- [x] Pass the design config to schema
- [x] Make theme passable as a UI option.
- [x] Add document
- [x] Merge PR: https://github.com/pdfme/pdfme/pull/322

`options.customTheme`

---

### Is your feature request related to a problem? Please describe.

The design of pdfme might not match the style of a user's app when using the default design. Therefore, we allow customization to suit the user's preferences and style.

### Describe the solution you'd like

- Allow changes to the accent color (currently it's light blue)
- Add class names to each UI component to allow customization from the outside.

### Describe alternatives you've considered

None, currently we can't customize style.

### Additional context

ref: https://pdfme.com/development-guide#v3-roadmap

hand-dot:
https://ant.design/docs/react/customize-theme

PR #322, opened 2023-11-16
Custom Design for UI#243

hand-dot:
https://github.com/pdfme/pdfme/issues/243


playground/src/Designer.tsx
```diff
@@ -36,7 +36,14 @@ function App() {
         designer.current = new Designer({
           domContainer: designerRef.current,
           template,
-          options: { font },
+          options: {
+            font,
+            customTheme: {
```
hand-dot, line 41: We can use like this.

packages/ui/src/theme.ts
hand-dot, line 1: user can overwrite based on this defaultTheme.

playground/src/Designer.tsx
```diff
@@ -36,7 +36,14 @@ function App() {
         designer.current = new Designer({
           domContainer: designerRef.current,
           template,
-          options: { font },
+          options: {
+            font,
+            theme: {
```
hand-dot, line 41: We can use like this.

Customize Design Token: https://ant.design/docs/react/customize-theme#customize-design-token

hand-dot:
Hey @peteward ,

If you have time, could you please check this PR?

packages/ui/src/components/Designer/Sidebar/index.tsx
```diff
@@ -60,47 +51,32 @@ const Sidebar = (props: SidebarProps) => {
...
-                borderRadius: 2,
-                cursor: 'pointer',
-                margin: '0 auto',
-                display: 'block',
-              }}
-              onClick={addSchema}
-            >
-              <strong style={{ color: '#fff' }}>{i18n('addNewField')}</strong>
-            </button>
+            <Button type="primary" style={{ fontWeight: 600 }} onClick={addSchema}>
```
peteward, line 77: do you need to override font weights here? shouldn't the primary button style take care of this?

peteward (approved):
This is excellent, very exciting!

Issue #107, opened 2022-11-10
Custom Label for UI


hand-dot:
ref: https://github.com/pdfme/pdfme/pull/323

---

Currently, changing the language settings translates some labels, but eventually, the aim is to make all labels translatable, and additionally add a setting to overwrite specific keys.

- [x] Remove unused labels from i18n.ts as well.
- [x] Make i18n overwritable from options
- [x] Retrieve existing schema labels from the labels in options
- [x] Add a tips to the documentation for overriding the official schema labels
- [x] Release
  - [x]  npm
  - [x]  website
  - [x]  playground

`options.labels`

---

ref https://github.com/pdfme/pdfme/discussions/98
ref https://github.com/pdfme/pdfme/pull/100
ref https://github.com/pdfme/pdfme/issues/265

MarioWork:
maybe this note can be removed from docs 
https://pdfme.com/docs/custom-ui
<img width="768" height="104" alt="Image" src="https://github.com/user-attachments/assets/6763416c-1e32-45c1-846f-a37ae31f8c79" />

PR #323, opened 2023-11-20
Custom label UI 107

hand-dot:
ref: https://github.com/pdfme/pdfme/issues/107


playground/src/Designer.tsx
```diff
@@ -38,6 +39,11 @@ function App() {
           template,
           options: {
             font,
+            lang,
+            labels: {
```
hand-dot, line 43: We can overwrite labels like this.

hand-dot:
Hey @peteward ,
Can you review it when you have time? 
CI is failing, but it's unrelated to this fix, so I'd like to address it in a separate PR.

packages/schemas/src/text/propPanel.ts
```diff
@@ -58,72 +58,92 @@ export const propPanel: PropPanel<TextSchema> = {
...
-        title: 'Text Align',
+        title: i18n('schemas.text.textAlign'),
         type: 'string',
         widget: 'select',
         props: {
           options: [
-            { label: 'Left', value: DEFAULT_ALIGNMENT },
-            { label: 'Center', value: ALIGN_CENTER },
-            { label: 'Right', value: ALIGN_RIGHT },
+            { label: i18n('schemas.left'), value: DEFAULT_ALIGNMENT },
```
peteward, line 87: separate issue, but we should probably have an `ALIGN_LEFT` constant

packages/ui/src/i18n.ts
```diff
@@ -22,17 +23,37 @@ const dictEn = {
     'Cannot commit the change because the number of items has been changed.',
   commitBulkUpdateFieldName: 'Commit Changes',
   bulkUpdateFieldName: 'Bulk update field names',
-  barColor: 'Bar Color',
-  textColor: 'Text Color',
-  bgColor: 'Background Color',
+  'schemas.textColor': 'Text Color',
+  'schemas.bgColor': 'Background Color',
```
peteward, line 27: should these be defined in the schema package, under the text/ folder?

peteward (approved):
Excellent.

If you are writing a custom schema plugin how do you define the translations for it which can be merged with the main translation list?

I can't see how you can do this currently, all of the translations are in [i18n.ts](https://github.com/pdfme/pdfme/pull/323/files#diff-b8ff8216939f858f3ddc502c04c6bf19e282d3d666487ec842a61cc1501adc5e).

Maybe as a next step we should support merging translation keys from within schema directories (hence my suggestion about moving some of the schema-related translation keys to the schema package)

hand-dot:
@peteward 
Thank you for reviewing.

When customizing labels that are not part of the official schema, users can add labels as shown below.

https://github.com/pdfme/pdfme/blob/a7217408b16286ea9c881cf28688812d021331b7/playground/src/FormAndViewer.tsx#L61

Forcing the addition of labels for each language under the schema package would be a significant overhead for plugin developers, so I have done it this way.

As you suggest, merging into the main translation list is also a good idea, but I thought it was overkill at this stage, so I implemented it minimally."

peteward:
ok great.

Yes, I saw that you can add your own labels - which is great.

But if you want to write a custom plugin for other people to use, then ideally they should be able to use it without having to pass labels through the designer. 

Something to consider for future.

Issue #326, opened 2023-11-21
Error [ERR_REQUIRE_ESM]: require() of ES Module


hafees20:
### Describe the bug

The issue is arising because the **@pdfme/generator** package has **"type": "module"** in its **package.json**. This means that the package is treated as an ECMAScript module (ESM), and when it's required in a file with "type": "commonjs", such as a typical Node.js application, you encounter the Error [ERR_REQUIRE_ESM].

Changing the **"type": "module"** to  **"type": "commonjs"** in the **package.json** file will temporarily fix the issue as per the log.

![Screenshot from 2023-11-21 16-44-05](https://github.com/pdfme/pdfme/assets/81296862/86c8a6bb-eedd-4944-b42c-ca13f5251b95)

### To Reproduce

When the package is required in a file with type:commonjs.

### Expected behavior

.

### Your Environment

```markdown
- pdfme package : @pdfme/generator and @pdfme/common and @pdf/schemas
- pdfme version : 3.1.1
- Operating system : Linux Mint 21.2
- Node.js version : v18.16.0
```

### Your Error Log

```shell
const generator_1 = require("@pdfme/generator");

Error [ERR_REQUIRE_ESM]: require() of ES Module

index.js is treated as an ES module file as it is a .js file whose nearest parent package.json contains "type": "module" which declares all .js files in that package scope as ES modules.
Instead rename index.js to end in .cjs, change the requiring code to use dynamic import() which is available in all CommonJS modules, or change "type": "module" to "type": "commonjs"
```

thiagotognoli:
Facing the same issue.

hafees20:
> Facing the same issue.

Check the temporary fix in my comment above

hand-dot:
Can you check pdfme@3.1.2 ?
https://github.com/pdfme/pdfme/releases/tag/3.1.2

chalkygames123:
Hi @hand-dot,

Just removing the `type` field is not a proper way to fix this issue because ESM projects will no longer be able to import this library. In short, while using ESM as the default module format with `"type": "module"`, you can selectively use CommonJS with the `.cjs` extension. I'll send a PR if you want.

PR #328, opened 2023-11-23
Bugfix module type package

thiagotognoli:
ref #326


hand-dot:
Fixed by https://github.com/pdfme/pdfme/pull/339

Issue #334, opened 2023-11-26
When switching from one scheme to another, the rotation value did not reset.


LeandroBanaszeski:
### Describe the bug

![image018](https://github.com/pdfme/pdfme/assets/45818324/2fbfbaea-574c-4705-8fcb-1773ab7cc8ff)

### To Reproduce

1 - Select any schema in the UI.
2 - Change the rotation to any value.
3 - Select another schema in the UI.

### Expected behavior

Rotation field should reset when changing schema, but it continues with the value of the last selected schema.

### Your Environment

```markdown
- pdfme package(@pdfme/generator or @pdfme/ui): @pdfme/ui
- pdfme version: 3.1.0
- Operating system: Windows 
- Node.js version or Browser name & version: 18.17.1
```

### Your Error Log

```shell
1 - Select any schema in the UI.
2 - Change the rotation to any value.
3 - Select another scheme in the UI using the SideBar.

Rotation field returns default
```

hand-dot:
this was fixed by bellow code

https://github.com/pdfme/pdfme/blob/cde073cb510e9dadf57bd259b343dddba5bcf3c7/packages/ui/src/components/Designer/Sidebar/DetailView/index.tsx#L68

please wait to merge and release it.

hand-dot:
Fixed: https://github.com/pdfme/pdfme/pull/338

PR #337, opened 2023-11-26
Reset rotation field when changing schema in UI

peteward:
Ref: https://github.com/pdfme/pdfme/issues/334

Opened on behalf of @LeandroBanaszeski

### DIFF
--- a/packages/ui/src/components/Designer/Canvas/index.tsx
+++ b/packages/ui/src/components/Designer/Canvas/index.tsx
@@ -305,6 +305,9 @@ const Canvas = (props: Props, ref: Ref<HTMLDivElement>) => {
           if (!isClick && removed.length > 0) {
             newActiveElements = activeElements.filter((ae) => !removed.includes(ae));
           }
+          if(selected.length === 1){
+            onEdit([])
+          }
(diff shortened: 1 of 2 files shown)


hand-dot:
@peteward 
Thank you.

peteward:
I think the issue here is due to setting form values [here](https://github.com/pdfme/pdfme/blob/main/packages/ui/src/components/Designer/Sidebar/DetailView/index.tsx#L60C1-L62).

We're setting the values for the widgets based on the active schema, but it's possible that the incoming data is missing a value (such as rotation, or a colour which can be unset) and so the previous stale value remains.

peteward:
@hand-dot I think this solution is a bit more efficient and cleaner at the appropriate point. Let me know what you think.

hand-dot:
LGTM!

hand-dot (approved):

hand-dot:
@peteward 

Sorry, but when I debugged it again, it turns out this is causing a different issue.

 For example, if the position is changed by input entry in the Edit Field, the focus is lost. 
 Also, the validation for the min and max of rotate is not working.

Issue #336, opened 2023-11-26
Add opacity option to all schemas


LeandroBanaszeski:
### Is your feature request related to a problem? Please describe.

There was a need to create a watermark and one way would be to use opacity in the schemas

### Describe the solution you'd like

When selecting a schema, I can change the opacity

### Describe alternatives you've considered

Add opacity option to all schemas

PR #338, opened 2023-11-26
Added opacity option to all schemes and fix rotation bugs

peteward:
Ref: https://github.com/pdfme/pdfme/issues/336

Opened on behalf of @LeandroBanaszeski


hand-dot:
@peteward 
I'd like to hear your opinion on this feature.

I think an input UI for a number range from 0 to 1 would be better since the slider UI seems a bit exaggerated. Also, similar to rotate, I believe we need controls such as disabling it when opacity is not specified in the plugin's propPanel.defaultSchema.

If you agree with the above, I can continue with the implementation on my end.

Personally, I don't think it's an absolutely necessary feature, but if it doesn't interfere with the designer's UI, I do think it's a nice feature to have.

peteward:
@hand-dot ,

Whilst not a high priority feature, it clearly has a use-case for @LeandroBanaszeski and it seems quite simple for both the UI and pdf-lib so I think we should add it.

I think the slider is OK, but not at full width, it takes up too much space. It also seems a to be visually clipped at the edges, and buggy when I put it at the end of a row - it causes the UI to refresh for me. So, perhaps just keep it simple with a number input.

Agree that we should make it optional 👍

packages/ui/src/components/Designer/Sidebar/DetailView/index.tsx
```diff
@@ -58,16 +58,23 @@ const DetailView = (
   }, [activeSchema, activeElements, pluginsRegistry, JSON.stringify(options)]);
 
   useEffect(() => {
-    form.resetFields();
-    form.setValues({ ...activeSchema });
+    const values: any = { ...activeSchema };
```
hand-dot, line 61: I know this code isn't very elegant. However, it was necessary to change the way position.x and position.y data are handled in order to create a three-column layout in form-render.

packages/ui/src/components/Designer/Sidebar/DetailView/index.tsx
```diff
@@ -93,34 +106,46 @@ Check this document: https://pdfme.com/docs/custom-schemas`);
...
-          options: Object.entries(pluginsRegistry).map(([label, value]) => ({
-            label,
-            value: value?.propPanel.defaultSchema.type,
-          })),
-        },
-      },
-      key: { title: i18n('fieldName'), type: 'string', widget: 'input' },
-      '-': { type: 'void', widget: 'Divider', cellSpan: 2 },
-      align: { title: i18n('align'), type: 'void', widget: 'AlignWidget', cellSpan: 2 },
-      position: {
```
hand-dot, line 106: In nested data, the settings for 'span' do not work as intended.

packages/ui/src/components/Designer/Sidebar/DetailView/index.tsx
```diff
@@ -58,16 +58,23 @@ const DetailView = (
   }, [activeSchema, activeElements, pluginsRegistry, JSON.stringify(options)]);
 
   useEffect(() => {
-    form.resetFields();
-    form.setValues({ ...activeSchema });
+    const values: any = { ...activeSchema };
+    values.x = values.position.x;
+    values.y = values.position.y;
+    delete values.position;
```
hand-dot, line 64: I know this code isn't very elegant. However, it was necessary to change the way position.x and position.y data are handled in order to create a three-column layout in form-render.

peteward (commented):
excellent! 

What was the fix for the rotation issue?

hand-dot:
Hey @peteward 

> I uploaded a version of the schema after removing the optional rotation field to trigger the bug.
Where should i check?

peteward:
no need to apologise, great work! I am sorry for breaking it :D 

I can confirm this fix with the local template I made

peteward:
Hi @hand-dot,

This appears to have introduced a new issue:


https://github.com/pdfme/pdfme/assets/7068515/af2808fe-2ced-4847-8a5d-d42512d33145

* Select a schema 
* Enter a value into one of the inputs
* Select a different input
* The schemas now keep selecting and unselecting when you click one of them.

packages/ui/src/components/Designer/Sidebar/DetailView/index.tsx
```diff
@@ -93,34 +114,46 @@ Check this document: https://pdfme.com/docs/custom-schemas`);
...
+        title: i18n('height'),
+        type: 'number',
+        widget: 'inputNumber',
+        required: true,
+        span: 8,
+      },
+      opacity: {
+        title: i18n('opacity'),
+        type: 'number',
+        widget: 'inputNumber',
```
peteward, line 152:
```suggestion
        widget: 'inputNumber',
        step: 0.1,
```

hand-dot:
@peteward 

Oh....., okay i'll check it! thank you!

packages/ui/src/components/Designer/Sidebar/DetailView/index.tsx
peteward: hmm this doesn't work, I will figure this out and update

hand-dot:
@peteward 
Reproducing the issue is not possible, is there something wrong with the procedure?


https://github.com/pdfme/pdfme/assets/24843808/004d7e51-a20f-4674-8900-546b1131608f

peteward:
I've done a fresh checkout, switched to this branch and built as before.
I'm seeing the same bouncing behaviour when I edit more than one input on a schema.

peteward:
Ok, I have figured this out with @steffancarrington's help!

* When you resize the screen, or the container size changes, the sidebar re-renders to the list view, even if you are on the detail view.
* When you increase opacity > 1 you get the red error message displayed.
* In my version of Chrome, this error message caused the container size to grow and create a scrollbar around the sidebar. This triggered a re-render, which closed the sidebar and started a bounce loop.
* When I updated to the latest Chrome this problem did not happen.

I think there is a bug here that container or screen size changes that trigger re-render should not lose the selected schema state, but it is not related to this PR. 

I will log this separately to look at.

peteward:
I cannot approve this PR because I opened it, but it's great :D 👍

hand-dot:
Hey, @peteward 

Thank you for taking the time to debug. I respect your thorough quality check.

OK, I'll still hold off on merging this.
However, I don't want the PR to become outdated and neglected, so please feel free to let me know if there's anything.
I will also do some debugging on my end.

peteward:
this is very strange...

I _think_ it is related to a LastPass browser extension that I have. 

What is strange is that just adding the `opacity` field definition in creates the problem and seems to activate LastPass on the form. If I comment this out, then I do not see LastPass running on these fields:

```
      opacity: {
        title: i18n('opacity'),
        type: 'number',
        widget: 'inputNumber',
        disabled: defaultSchema?.opacity === undefined,
        dataLpignore: true,
        max: 1,
        min: 0,
        span: 8,
      },
```

->

Screenshot 2023-12-02 at 16 26 48

I don't know what it is about this field that seems to trigger Lastpass. 

I will try to look at what is losing the schema selection when the screen resizes, I think this will resolve the issue.

hand-dot:
@peteward Thank you!

Issue #347, opened 2023-12-04
Playground React.StrictMode causing duplicate renders which lead to triggering template change


peteward:
### Describe the bug

React.StrictMode that is used within the playground side is leading to some strange side effects.

I have not manage to fully understand the side effects, but I am listing my findings here and proposing we remove StrictMode for now unless a better solution can be found.

The UI app creates an instance of `BaseUIClass`, which sets up a ResizeObserver, which in turn re-renders when an event is received:

```
  private readonly setSize = debounce(() => {
    if (!this.domContainer) throw Error(DESTROYED_ERR_MSG);
    this.size = {
      height: this.domContainer.clientHeight || window.innerHeight,
      width: this.domContainer.clientWidth || window.innerWidth,
    };
    this.render();
  }, 100);
```

The Designer class which extends this class renders the component including:

```
        <DesignerComponent
          template={this.template}
```

Inside this component there is a useEffect which updates if the template changes:

```
  const updateTemplate = useCallback(async (newTemplate: Template) => {
    const sl = await templateSchemas2SchemasList(newTemplate);
    setSchemasList(sl);
    onEditEnd();
    setPageCursor(0);
    if (mainRef.current?.scroll) {
      mainRef.current.scroll({ top: 0, behavior: 'smooth' });
    }
  }, []);

  useEffect(() => {
    updateTemplate(template);
  }, [template, updateTemplate]);
```

2 things seem to be happening with StrictMode:

1. The component is being rendered twice. However, these appear to be isolated, which doesn't appear to be the problem.
2. The above `useEffect` is seemingly being triggered with StrictMode when it shouldn't (it seems to think `template` has changed, despite debugging that it is identical.

For this second point, it's possibly due to:

> When Strict Mode is on, React will also run one extra setup+cleanup cycle in development for every Effect. This may feel surprising, but it helps reveal subtle bugs that are hard to catch manually.

https://react.dev/reference/react/StrictMode

It's possible that there is a bug in the code, but it doesn't seem obvious to me. 

Unless this issue can be resolved I suggest we remove `<React.StrictMode>` from the playground.

This is the problem that impacts #

### To Reproduce

In the playground (not the demo site):
* select a schema
* resize the window
* schema selection will be lost and a heavy re-render will happen.

### Expected behavior

Window should resize without losing the scheam in focus.

### Your Environment

```markdown
- pdfme package: @pdfme/ui
- pdfme version: latest main
- Operating system: OSX
- Node.js version: 18
[...]

PR #348, opened 2023-12-04
remove React.StrictMode in playground to avoid full rerender on screen resize

peteward:
See #347, but there might be a better way...


hand-dot:
Hey @peteward ,

Thank you for the PR!

However, I think it's better to have React.StrictMode during development, so I can't merge this approach...

peteward:
I think I have an appropriate solution for this now. I will open another PR soon.

PR #356, opened 2023-12-05
Refactor to remove inefficient useEffect with side effects

### LINKED ISSUE #347: Playground React.StrictMode causing duplicate renders which lead to triggering template change

peteward:
fixes #347

It took a long time to track down the issues that were being exposed by `<React.StrictMode>`. 
After referring to the following [You might not need an Effect guide here](https://react.dev/learn/you-might-not-need-an-effect#adjusting-some-state-when-a-prop-changes), it explained that we have used one of the patterns that they recommend against in a few places. Not only is it less efficient because of additional re-renders, but in our case the `useEffect` in the playgrounds were leading to duplicate calls with alternative template objects which resulted in an additional re-render and a deselection of the schema. 

**Before:**
* loses focus on resize of the window
* resize causes the canvas area to flash as schema is unnecessarily re-rendered
* can get caught in a select/unselect state if anything causes the sidebar size to increase

https://github.com/pdfme/pdfme/assets/7068515/45d29ed4-aac9-4816-88ea-4b2aa2cebdf1

**After:**
* fixes deselection issues
* much smoother re-rendering on resize

https://github.com/pdfme/pdfme/assets/7068515/669fe337-2a6a-41fa-9869-7521d97a63cf

I have not tried to refactor more widely than the areas immediately impacted, but it's worth considering whether we need a useEffect in more areas in future.

### DIFF
--- a/packages/ui/src/components/Designer/index.tsx
+++ b/packages/ui/src/components/Designer/index.tsx
@@ -1,4 +1,4 @@
-import React, { useRef, useState, useEffect, useContext, useCallback } from 'react';
+import React, {useRef, useState, useEffect, useContext, useCallback, useMemo} from 'react';
 import {
   ZOOM,
   Template,
@@ -49,6 +49,7 @@ const TemplateEditor = ({
   const [pageCursor, setPageCursor] = useState(0);
   const [zoomLevel, setZoomLevel] = useState(1);
   const [sidebarOpen, setSidebarOpen] = useState(true);
+  const [prevTemplate, setPrevTemplate] = useState<Template | null>(null);
 
   const { backgrounds, pageSizes, scale, error } = useUIPreProcessor({ template, size, zoomLevel });
 
@@ -148,10 +149,6 @@ const TemplateEditor = ({
     }
   }, []);
 
-  useEffect(() => {
-    updateTemplate(template);
-  }, [template, updateTemplate]);
-
   const addSchema = () => {
     const propPanel = (Object.values(pluginsRegistry)[0] as Plugin<Schema>)?.propPanel;
 
@@ -182,6 +179,11 @@ Check this document: https://pdfme.com/docs/custom-schemas`);
   const onChangeHoveringSchemaId = (id: string | null) => {
     setHoveringSchemaId(id);
   };
+
+  if (prevTemplate !== template) {
+    setPrevTemplate(template);
+    void updateTemplate(template);
+  }
 
   const sizeExcSidebar = {
     width: sidebarOpen ? size.width - SIDEBAR_WIDTH : size.width,
--- a/packages/ui/src/components/Designer/index.tsx
+++ b/packages/ui/src/components/Designer/index.tsx
@@ -1,4 +1,4 @@
-import React, {useRef, useState, useEffect, useContext, useCallback, useMemo} from 'react';
+import React, { useRef, useState, useContext, useCallback } from 'react';
 import {
   ZOOM,
   Template,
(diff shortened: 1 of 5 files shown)


packages/ui/src/components/Designer/index.tsx
```diff
@@ -1,4 +1,4 @@
-import React, { useRef, useState, useEffect, useContext, useCallback } from 'react';
+import React, {useRef, useState, useEffect, useContext, useCallback, useMemo} from 'react';
```
peteward, line 1: remove

playground/src/FormAndViewer.tsx
```diff
@@ -1,6 +1,6 @@
 import { useEffect, useRef, useState } from "react";
 import { Template, checkTemplate } from "@pdfme/common";
-import { Form, Viewer } from "@pdfme/ui";
+import {Designer, Form, Viewer} from "@pdfme/ui";
```
peteward, line 3: space

hand-dot (approved):
@peteward 
Great....!

As you pointed out, I think there are problems elsewhere too. This is what I have done, but I believe we need to improve the current situation where eslint errors and warnings are being ignored.

Issue #357, opened 2023-12-06
Dynamic font sizing no longer truly dynamic


peteward:
### Describe the bug

Since the implementation of custom schema, and the change to non-react uiRender files, we have lost the benefit of optimized React rendered components.

As a result, schema rendering now replaces the entire content when re-rendered, and to avoid how heavy this is we are now only re-rendering with an `onChange` event for when the editing has finished.

This is a regression in user experience for the text schema, especially when using dynamic text sizing. Previously this would adjust as you type to fit the available space (as per Google Docs, etc):

https://github.com/pdfme/pdfme/assets/7068515/4614533d-9552-4fd9-b0aa-31b727fd1c28

This no longer happens, the sizing of text is only applied once editing has finished and you leave the editor window:

https://github.com/pdfme/pdfme/assets/7068515/970d5705-804b-4157-82ec-26186e936414

I have tried changing the event listener to be onKeyPress, but without the use of React this is far too janky because it rerenders and calculates every aspect of the component on every keypress. There were delays of up to 20 seconds to update the display when I typed for a few seconds with lots of text. There were also issues with `focus()` in doing this if you are typing in the middle of the textbox it will move focus to the end.

### To Reproduce

Type into a text schema

### Expected behavior

Ideally, we would be able to leverage React to support some dynamic updates to the visual display of the schema.

### Your Environment

```markdown
- pdfme package: @pdfme/ui
- pdfme version: latest main
- Operating system: OSX
- Node.js version or Browser name & version: Chrome latest
```

### Your Error Log

```shell
-
```

### Additional context

I don't know if this is easy to resolve.

Would it be possible to allow React components as an alternative to the basic implementations we have of the uiRender file? 

It would be shame to lose the user experience that we used to have.

hand-dot:
Hey @peteward ,
> Would it be possible to allow React components as an alternative to the basic implementations we have of the uiRender file?

I don't think it's a good idea to allow external React components. We would need to ensure that the React version used by pdfme matches the user's runtime, which complicates things. I believe it's simpler to have it written in vanilla JavaScript.

Couldn't we revert to the original UX by re-executing `calculateDynamicFontSize` within the callback of `onChange` mentioned below?
https://github.com/pdfme/pdfme/blob/67439652affbdd70b187a4583283c2cb3bb8b311/packages/schemas/src/text/uiRender.ts#L143

hand-dot:
[...]

PR #359, opened 2023-12-06
Use contenteditable as opposed to textarea for text fields

### LINKED ISSUE #345: Form ignores settings for verticalAlignment
### Describe the bug

When a field has `verticalAlignment` set to `middle` for example, the text is not centered vertically when the template is loaded as `Form`. It renderes correctly in `Designer`, `Viewer` and when a PDF is generated.

### To Reproduce

1. Start the Playground
2. Set the vertical alignment to `middle`
3. Save the template and switch to the `Form` view

### Expected behavior

Apply the same settings as in `Designer` and `Viewer`.

### Your Environment

```markdown
- pdfme version: `main` branch
> $ git log -n 1 --pretty=format:"%H"
> 3258e441b18c7c93cdf2a715a904213d9b97fc6c
- Operating system: Ubuntu 23.10
- Node.js version or Browser name & version: Node v20.10, Chrome 119.0
```

### Your Error Log

```shell
There is no error log, but here is a screencast below:
```

### Additional context

[Screencast from 2023-12-01 23-04-24.webm](https://github.com/pdfme/pdfme/assets/227762/324403c2-b079-45ee-b715-5be44dd551a7)

peteward:
fixes #345 - for text fields to respect vertical alignment while in editing mode.

Another positive point about using contenteditable is that the style is the same for when you are in edit mode and when you are not. You do not have to worry about 'adjusting' font positions in the different cases.

However, I think we might have lost some positioning adjustments with the custom schema work, so I will revisit that in a separate issue.

Example of change:

https://github.com/pdfme/pdfme/assets/7068515/cce219e5-23c2-4562-89a0-0dba8b51fe74


packages/schemas/src/text/uiRender.ts
```diff
@@ -119,45 +120,56 @@ export const uiRender = async (arg: UIRenderProps<TextSchema>) => {
...
+      const placeholderStyle = document.createElement('style');
+      placeholderStyle.textContent = `
+        [contenteditable=true]:empty:before {
+          content: attr(placeholder);
+          pointer-events: none;
+          display: block;
+        }
+      `;
+      container.appendChild(placeholderStyle);
+    }
```
peteward, line 156: There was an implementation for `placeholder` before, so I have added this to maintain the behaviour, although I'm not sure if it's possible to use it?

peteward:
hmm tests all pass locally which is strange.. 
```
> @pdfme/ui@0.0.0 test
> jest

 PASS  __tests__/helper.test.ts
 PASS  __tests__/components/Designer.test.tsx
 PASS  __tests__/components/Preview.test.tsx

Test Suites: 3 passed, 3 total
Tests:       13 passed, 13 total
Snapshots:   3 passed, 3 total
Time:        4.143 s
Ran all test suites.
```

I will check again tomorrow

packages/schemas/src/text/uiRender.ts
hand-dot: I checked, but the placeholder disappears. We need to think of a way to display the placeholder in the div.

hand-dot:
Hey @peteward 

Can you check bellow changes?

- [Fix Dynamic font sizing no longer truly dynamic #357
](https://github.com/pdfme/pdfme/pull/359/commits/88868c95028bfdc2e5478a9a55f7aed8b3b66157)
- [Fix placeholder](https://github.com/pdfme/pdfme/pull/359/commits/179a895afc30d5ae3ef46219a5ded2e8ec072515)

https://github.com/pdfme/pdfme/assets/24843808/4ae5ffef-d18a-4ef4-b057-94fac8c857dc

packages/schemas/src/text/uiRender.ts
```diff
@@ -118,45 +122,66 @@ export const uiRender = async (arg: UIRenderProps<TextSchema>) => {
...
-    container.appendChild(textarea);
+    textBlock.contentEditable = 'plaintext-only';
+    textBlock.tabIndex = tabIndex || 0;
+    textBlock.innerText = value;
+    textBlock.addEventListener('blur', (e: Event) => {
+      onChange && onChange((e.target as HTMLDivElement).innerText);
+      stopEditing && stopEditing();
+    });
+
+    textBlock.addEventListener('keypress', () => {
```
peteward, line 145: ah of course, you can just do it this way 🤦  

I will test this locally very soon

peteward:
This is a great improvement. A couple of issues to be looked at though:

1. The topAdj and bottomAdj need to be recalculated when font size changes, so need to also be called on the keyUp handler.
2. It's still a lot jankier than it used to be.

I'm happy to look into both of these issues over the weekend.

peteward:
Updated:
* use `keyup` instead of `keypress` so that it triggers when backspace and delete are pressed
* recalc the dynamic font size based on the previous calculated font size (rather than the base one) as this considerably reduces the number of calculation iterations required.
* recalculate the bottom and top adjustments along with font size.

@hand-dot I think this one is good to go now! Can you confirm?

hand-dot:
@peteward 
Cool! I'll check it and If it okay i gonna merge this PR.

Thank you

hand-dot (approved):
LGTM!
Really nice!!

hand-dot:
Released V3.1.4

Issue #354, opened 2023-12-05
Validating the Schema in the Designer


peteward:
### Is your feature request related to a problem? Please describe.

We embed pdfme in an application that allows end-users to design their own layout. Once they have finished then can either preview or save the schema for use against real data.

Sometimes they may end up in an invalid state:
![Screenshot 2023-12-05 at 09 09 25](https://github.com/pdfme/pdfme/assets/7068515/1e0eb05a-a786-4e15-b664-7c12adc47129)
![Screenshot 2023-12-05 at 09 09 59](https://github.com/pdfme/pdfme/assets/7068515/d09273e1-440f-44ec-b796-29e1b6c90577)

but there is no API that you can check built into pdfme (Designer) that will tell you about the validity of the schema in its current state.

### Describe the solution you'd like

I think it would be great if we could call a function such as `designer.isSchemaValid()` to tell externally whether the schema is valid.  Since we have the capability to embed validation rules into the form-render definitions, it makes sense to pdfme to leverage this. 

### Describe alternatives you've considered

Doing validation externally to pdfme, but this feels like the wrong approach.

hand-dot:
Can't the following code achieve the same thing?

```ts
import { checkTemplate } from "@pdfme/common";

if (checkTemplate(designer.getTemplate())) {
  console.log("valid");
} else {
  console.log("invalid");
}
```

However, I think your suggested `designer.isSchemaValid()` is more user-friendly than the code above.

If you decide to add it, I would like to accept the Pull Request.

peteward:
`checkTemplate` is good, but it doesn't cover many of the attributes within a schema.

It only really covers the base attributes:

```
  .object({
    type: z.string(),
    position: z.object({ x: z.number(), y: z.number() }),
    width: z.number(),
    height: z.number(),
    rotate: z.number().optional(),
    opacity: z.number().optional(),
  })
``` 

so for a test schema it won't validate the background color is a valid hex code.
It also will check for a number on `width` but there's no validation that it's a positive number.

These validation rules we are starting to implement into the UI propPanel JSON. My idea was to use this to drive validation for checking the whole schema. However, this will only work within the UI, so won't be much use for the generator. But it will be useful for the standard workflow where the designer/form will be defining the schema to be used in the generator.

If I have time I will try it tomorrow and share my idea.

hand-dot:
form-render instance has `validateFields` methods.
https://xrender.fun/form-render-mobile/api#forminstance

[...]

Issue #342, opened 2023-11-27
Error when trying to preview PDF if rotation is null


LeandroBanaszeski:
### Describe the bug

![image018](https://github.com/pdfme/pdfme/assets/45818324/758d4d2e-017f-404c-86da-8a93c755d248)

Error when trying to preview PDF if rotation is null

In the model it is currently optional, but it does not accept null values, perhaps an option would be to leave it with nullable in the model,
Another option would be, when deleting the value from the rotation input field, it should be removed from Template.

### To Reproduce

1 - Select a schema in the UI
2 - Add value to rotation
3 - Click on the rotation input field and delete the value, leaving it blank
4 - Try previewing the PDF

### Expected behavior

Preview should open correctly

### Your Environment

```markdown
- pdfme package(@pdfme/generator or @pdfme/ui): @pdfme/ui
- pdfme version: 3 latest
- Operating system: Windows
- Node.js version or Browser name & version: 18.17.1
```

### Your Error Log

```shell
Uncaught (in promise) Error: [@pdfme/common] Invalid argument:
--------------------------
ERROR POSITION: template.schemas.0.photo.rotate
ERROR MESSAGE: Expected number, received null
```

peteward:
Thanks for reporting this @LeandroBanaszeski !

What do you think about the options @hand-dot ?

Either add `.nullable()` to the schema definition here (and test the impact):

```
export const Schema = z
  .object({
    type: z.string(),
    position: z.object({ x: z.number(), y: z.number() }),
    width: z.number(),
    height: z.number(),
    rotate: z.number().optional(),
  })
```
 
Or when a value is deleted in an input in the UI we remove from the schema?

hand-dot:
@peteward 
This was fixed by bellow code.

https://github.com/pdfme/pdfme/blob/cde073cb510e9dadf57bd259b343dddba5bcf3c7/packages/ui/src/components/Designer/Sidebar/DetailView/index.tsx#L80

PR #367, opened 2023-12-14
Fix to allow zero values for form fields

### LINKED ISSUE #366: Deleting a core value from the UI (e.g. X or Y pos), or setting to 0 will remove from the schema
### Describe the bug

If you have a 0 value for some parameters they will be removed from the resulting schema, which then becomes invalid and won't load

### To Reproduce

* Create a text schema
* Delete the 'Y' value for it in the UI
* Download the template for the Schema
* The y value will be missing, e.g.:

```
{"schemas":[{"name":{"type":"text","position":{"x":24.77},"width":77.77,"height":38.01,....
```

### Expected behavior

You should not be able to fully delete a numeric required value.
Any attempt to delete it should 0 the value, and this should be retained when the schema is saved.

### Your Environment

```markdown
- pdfme package: @pdfme/ui
- pdfme version: 3.1.5
- Operating system: Mac OSC
```

### Your Error Log

```shell
On loading a broken schema:


Error: [@pdfme/common] Invalid argument:
--------------------------
ERROR POSITION: template.schemas.0.name copy.position.y
ERROR MESSAGE: Required
--------------------------
    at L6 (index.js:159:10)
    at zbe (index.js:161:314)
    at new ENe (index.js:266:9838)
```

You will also get a NaN error for applying padding to the text.
```


### Additional context

_No response_

peteward:
Fixes #366 

I can't see any negative impact of this change. 

It's still technically possible to break a schema by forcibly deleting a required value and ignoring the red form error message, but it's a lot better than having the value removed from the schema entirely if it's value is 0.

You can still remove values such as rotation from the schema, but by deleting the values not by setting them as 0. If we wanted to do this we would need to check if these fields are optional or not then remove them. This is more complex to implement.

https://github.com/pdfme/pdfme/assets/7068515/b3350f0c-111d-48ec-8c1a-5a100d4ee4ba

### DIFF
--- a/packages/ui/src/components/Designer/Sidebar/DetailView/index.tsx
+++ b/packages/ui/src/components/Designer/Sidebar/DetailView/index.tsx
@@ -77,7 +77,8 @@ const DetailView = (
     for (let key in newSchema) {
       if (['id', 'data'].includes(key)) continue;
       if (newSchema[key] !== (activeSchema as any)[key]) {
-        const value = newSchema[key];
+        let value = newSchema[key];
+        if (value === null && ['rotate', 'opacity'].includes(key)) value = undefined;
 
         // [position] Return the flattened position to its original form.
         if (key === 'x') key = 'position.x';
--- a/packages/ui/src/components/Designer/Sidebar/DetailView/index.tsx
+++ b/packages/ui/src/components/Designer/Sidebar/DetailView/index.tsx
@@ -78,6 +78,7 @@ const DetailView = (
       if (['id', 'data'].includes(key)) continue;
       if (newSchema[key] !== (activeSchema as any)[key]) {
         let value = newSchema[key];
+        // FIXME memo: https://github.com/pdfme/pdfme/pull/367#issuecomment-1857468274
         if (value === null && ['rotate', 'opacity'].includes(key)) value = undefined;
 
         // [position] Return the flattened position to its original form.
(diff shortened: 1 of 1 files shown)


hand-dot:
Hey @peteward ,
I checked. 
I added this code a few weeks ago to solve [this problem](https://github.com/pdfme/pdfme/issues/342).
https://github.com/pdfme/pdfme/blob/main/packages/ui/src/components/Designer/Sidebar/DetailView/index.tsx#L80

However, with this PR,  [the problem](https://github.com/pdfme/pdfme/issues/342) seems to have reoccurred.

https://github.com/pdfme/pdfme/assets/24843808/decefe70-249a-4578-9e3a-b7d988b0637b


https://github.com/pdfme/pdfme/issues/366#issuecomment-1856185506
You are right.  My last bug fix is not perfect.

While it's not the most elegant solution, hardcoding to assign `undefined` to the `opacity` and `rotate` values if they are `null` could be a feasible workaround. What do you think?

peteward:
Ok, I've updated as you have suggested.

Longer-term this needs some thought. Ideally we should be able to use a definition to check and enforce an appropriate value. I tried pulling in the Zod definition but it was getting messy, and it only describes the core shared attributes of schema, not the individual ones.  I also tried getting the propPanel from the schema and trying to check if the field was `required` or not, but that wasn't easy and doesn't necessarily align to the problem we have here.


 Also, it's still possible to get into a bad state because there's no enforcement of values:

https://github.com/pdfme/pdfme/assets/7068515/1bcc9a14-cd71-4aee-9c7b-8ee88162af4c

I think we do need to make the [validation state of the template more easily available](https://github.com/pdfme/pdfme/issues/354). At least that would help with the above issue.

hand-dot (approved):
We have another issue, but we are definitely making progress. 

Let's merge this!

hand-dot:
Please ignore failing test.
also i create new issue about often fail CI.
https://github.com/pdfme/pdfme/issues/370

peteward:
Good teamwork!!

hand-dot:
I know! we are good team👍

PR #369, opened 2023-12-15
Fix \r\n line breaks in generator

peteward:
The precedence of this regex was wrong and it would lead to double line breaks for `\r\n`.  You don't see this when working with the UI because browsers will convert `\r\n` to just `\n`, but you see it when data is passed directly to the generator.

example:

```
const inputs = [{ a: "Dear Pete,\r\nI hope you like this.\r\nLove Mum."}];
```

before:
Screenshot 2023-12-15 at 13 50 54

after:
Screenshot 2023-12-15 at 16 28 04

### DIFF
--- a/packages/schemas/src/text/pdfRender.ts
+++ b/packages/schemas/src/text/pdfRender.ts
@@ -162,7 +162,7 @@ export const pdfRender = async (arg: PDFRenderProps<TextSchema>) => {
   };
 
   let lines: string[] = [];
-  value.split(/\r|\n|\r\n/g).forEach((line: string) => {
+  value.split(/\r\n|\r|\n/g).forEach((line: string) => {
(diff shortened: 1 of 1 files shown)


packages/schemas/src/text/pdfRender.ts
```diff
@@ -162,7 +162,7 @@ export const pdfRender = async (arg: PDFRenderProps<TextSchema>) => {
   };
 
   let lines: string[] = [];
-  value.split(/\r|\n|\r\n/g).forEach((line: string) => {
+  value.split(/\r\n|\r|\n/g).forEach((line: string) => {
```
hand-dot, line 165: More robust pattern 👍

hand-dot (approved):
LGTM!

Issue #331, opened 2023-11-24
Add Ellipse Schema


hand-dot:
ref: https://pdf-lib.js.org/docs/api/classes/pdfpage#drawellipse

hand-dot:
ref: https://github.com/pdfme/pdfme/pull/375

Issue #330, opened 2023-11-24
Add Rectangle Schema


hand-dot:
ref: https://pdf-lib.js.org/docs/api/classes/pdfpage#drawrectangle

hand-dot:
ref: ref: https://github.com/pdfme/pdfme/pull/375

PR #375, opened 2023-12-17
Add Ellipse and Rectangle Schema

hand-dot:
ref: 
- #331
- #330


packages/schemas/src/shapes/rectAndEllipse.ts
hand-dot, line 1: The main part of this PR is...

hand-dot:
@peteward can you check this PR?

hand-dot:
memo: to fix conflict

https://github.com/pdfme/pdfme/pull/380/files#diff-6686759c7bc6ba95211b21e6b5043a8496afd228513b01a77b4c67834c82fd8b

peteward (approved):
very cool.

Question - if you want to write a custom plugin that extends one that already exists, is it harder with this change to consolidate the separate files into a single `Plugin` definition?

For example, if I want to have another shape plugin and I want to use the same propPanel as your `shape` one, is it still easy to do this without duplicating the code?

hand-dot:
@peteward 

> Question - if you want to write a custom plugin that extends one that already exists, is it harder with this change to consolidate the separate files into a single Plugin definition?
> For example, if I want to have another shape plugin and I want to use the same propPanel as your shape one, is it still easy to do this without duplicating the code?

we can extend like this.


```ts
const yourShape = {
  ui: () => {},
  pdf: () => {},
  propPanel: {
    ...ellipse.propPanel,
    defaultSchema: {
      ...ellipse.propPanel.defaultSchema,
      type: 'yourShape',
    },
  },
};
```

PR #376, opened 2023-12-17
Tune the integration test speed checks

peteward:
After checking the times locally for all tests, I've consolidated the longer-running tests into their own file, with an extra 0.5 seconds allow to run.

I have also reduced the time required from up to 1.999 seconds to 1.5 seconds for some of the other tests (which should be much faster to run)

### DIFF
--- a/packages/generator/__tests__/assets/templates/index.ts
+++ b/packages/generator/__tests__/assets/templates/index.ts
@@ -69,6 +69,7 @@ export const envelope = {
 
 export const barcode = {
   Aone31555QRコード,
+  Aone31553QRコード,
   Aone72230JANコード短縮,
   Aone72230JANコード標準,
   barcodes,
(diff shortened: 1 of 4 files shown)


packages/generator/__tests__/assets/templates/index.ts
```diff
@@ -87,7 +86,9 @@ export const business = {
   労働条件通知書,
 };
 
-export const textType = {
+// These tests are slower, so we allow more time for them to pass
+export const slowerAndTextType = {
+  Aone31553QRコード,
```
peteward, line 91: I know this is now not in the suite with the rest of the barcodes, but it's running time was almost the same as the verticalAlignment ones, so it is also quite slow to run compared to the rest of the barcode ones

hand-dot (approved):
@peteward  Thank you!
LGTM!

PR #379, opened 2023-12-18
Make barcode text entry more readable

### LINKED ISSUE #377: QR Codes font color is diffcult to read
### Describe the bug

This affects the editors UI:

Screenshot 2023-12-18 at 10 34 53

Since the default for QR codes is black on white, black text is bad for readability

### To Reproduce

* Add a QR code, edit it

### Expected behavior

* you should be able to easily read a QR code

### Your Environment

```markdown
- pdfme package: ui
```

### Your Error Log

```shell
-
```

### Additional context

I think using the main theme colour for the text is probably a good starting point.
It's possible that this could also clash with barcodes, but it's less likely than the common case of black on white.

peteward:
fixes #377 

Before:
Screenshot 2023-12-18 at 14 26 25

After:
Screenshot 2023-12-18 at 14 25 35


hand-dot (approved):
Nice!

PR #383, opened 2023-12-19
Add tests for hex2RgbColor

peteward:
I was debugging an issue, it wasn't in this code but I wrote tests to prove it anyway :D


hand-dot (approved):
Love it!

Issue #385, opened 2023-12-21
Add Readonly SVG


hand-dot:

hand-dot:
implemented by https://github.com/pdfme/pdfme/pull/386

Issue #373, opened 2023-12-16
Add Readonly Image


hand-dot:

hand-dot:
implemented by https://github.com/pdfme/pdfme/pull/386

Issue #372, opened 2023-12-16
Add Readonly Text


hand-dot:

hand-dot:
implemented by https://github.com/pdfme/pdfme/pull/386

PR #386, opened 2023-12-21
Add ReadOnly Schemas

hand-dot:
https://github.com/pdfme/pdfme/assets/24843808/559ee9d9-9b60-4d5f-889c-91bfef362a65

- [Add Readonly SVG#385](https://github.com/pdfme/pdfme/issues/385)
- [Add Readonly Image#373](https://github.com/pdfme/pdfme/issues/373)
- [Add Readonly Text#372](https://github.com/pdfme/pdfme/issues/372)

---

TODO
- [x] Fix error when clicking on schema in UI development environment created by npm run dev
- [x] Remove TODO comment
- [x] Test
  - [x] Generate PDF with SVG schema will fail on nodejs env, we need to try again after pull latest cantoo-scribe/pdf-lib code to pdfme/pdf-lib -> Fixed by this great PR!! https://github.com/cantoo-scribe/pdf-lib/pull/42


hand-dot:
@peteward 
The build is failing in CI, but the implementation is already complete, so could you please review it?

I added a template named pet.json in packages/generator/__tests__/assets/templates/pet.json. This template is designed using a readOnly schema instead of basePdf.

https://github.com/pdfme/pdfme/assets/24843808/7d043ab0-784b-4d58-ad7e-445973cd608b

Since the design can be completed within pdfme, it should be possible to support schemas like dynamic tables, where the height changes depending on the data, in the future.

peteward (commented):
Amazing progress!

I have some thoughts about design from my own experience.

This implementation creates a division between Editable and Read Only schemas. They are now either one or the other. I'm wondering if we can make it more flexible to consider a mixture of static and dynamic content. 

In our case, we still work from a fork where we allow users to type in a mixture of static text and dynamic, e.g.:

```
Hi {{name}}, we hope you like your order!
```

In this fork we have an attribute named `schema.content` which contains the above text and we pre-process this to substitute values into the placeholders. We have some modifications to the editor code to use `content` rather than `value`, similar to how you have used `readOnlyValue`

I wonder if an attribute name such as `content` might be more generic and flexible for supporting some input content that is separate from the `value` for other plugin developers. 

I'm also wondering from a UX experience if there should be different schemas just because it's Read Only.  It might be clearer to have this as an option that you can toggle in the details panel. Or it might not, maybe it needs to be tried 🤔 


Anyway, these are just ideas. This PR looks great. I'm just starting work on my plugin now so I might suggest an iteration in future :)

packages/ui/src/components/Designer/index.tsx
```diff
@@ -113,6 +113,8 @@ const TemplateEditor = ({
           )?.propPanel;
           set(tgt, 'data', propPanel?.defaultValue || '');
           Object.assign(tgt, propPanel?.defaultSchema || {});
+        } else if (key === 'data' && tgt.readOnly) {
```
peteward, line 116: instead of this, would it be better to provide an alternative function to be passed into the renderers?

We have this `onChange` function for all schemas which represents when the `data` changes. 
For a flexible plugin API we could add a `onCustomAttributeChange` function, which allows the plugin to trigger a change on whichever attribute they like?

peteward: (I need something like this for my plugin :D )

hand-dot:
@peteward 

Thank you for the insightful comment.

When I first considered the design for ReadOnly, I initially had it switchable via a checkbox in the DetailView. However, when creating Line, Rectangle, and Ellipse, these are always ReadOnly, so I separated them into different schemas as ReadOnly and non-ReadOnly, as it is now.

In fact, with some shapes and the newly added Text, Image, and SVG, it seems unlikely that users would often add ReadOnly schemas, right? Considering this, I think enforcing a ReadOnly setting might increase the workload when creating new plugins.

I'm designing pdfme to be as uncomplicated as possible. While I may not fully understand your use case, can't the implementation be done by extending the Text schema without using the ReadOnly schema?

Would you like to create the plugin your team desires after merging this PR? If there are clear specifications and goals documented, I will support as much as I can.

peteward:
I will push up my branch now to show you what I am doing, it might make more sense then.

peteward:
@hand-dot your argument for Read Only as a schema type makes sense 👍  

see my [WIP plugin code](https://github.com/pdfme/pdfme/pull/388) to understand what I am trying to achieve and see if there is any overlap with your data structure.

hand-dot:
@peteward 
Okay I'm checking your new WIP PR.

Can i merge this PR?
I do not intend to release this PR until it resolves the issues with your use case, even after merging. It's also possible to make readOnlyValue into content.

peteward:
Of course!

PR #388, opened 2023-12-23
Multi-variable Text Plugin

peteward:
This plugin is functionally ready for review.  The behaviour is shown below:

https://github.com/pdfme/pdfme/assets/7068515/84d5164e-da7b-47e6-be99-15006aba58bd

The idea (which you suggested a while ago :) ) is to store the variables as a JSON string, but these become the input  (`data`) and what is normally data in other schemas becomes `content`.

This plugin will allow n variables to be added to a single text field, where n can be >= 0 (so this will also achieve a read-only schema). These variables are automatically detected as you type, and done so in an optimally efficient way that is usable even with dynamic font sizing enabled, as shown in the video above.

### TODO:

- [ ] Add tests where applicable


packages/schemas/src/text/propPanel.ts
```diff
@@ -49,6 +49,37 @@ const UseDynamicFontSize = (props: PropPanelWidgetProps) => {
   rootElement.appendChild(label);
 };
 
+// This will move into the plugin, I was just testing it here...
```
peteward, line 52: There will be no changes to this file, I can extend it from within the plugin

packages/ui/src/components/Designer/Canvas/index.tsx
```diff
@@ -381,6 +381,9 @@ const Canvas = (props: Props, ref: Ref<HTMLDivElement>) => {
             onChange={(value) => {
               changeSchemas([{ key: 'data', value, schemaId: schema.id }]);
             }}
+            onCustomAttributeChange={(key, value) => {
```
peteward, line 384: this gives flexibility to allow a ui renderer to make changes to different schema attributes, in my case `content`

packages/ui/src/components/Renderer.tsx
```diff
@@ -54,6 +55,13 @@ const Renderer = (props: RendererProps) => {
   const ref = useRef<HTMLDivElement>(null);
   const _cache = useRef<Map<any, any>>(new Map());
 
+  const flatSchemaWithoutData = JSON.stringify(schema, (key, value) => {
```
peteward, line 58: this was an important change to allow the propPanel to be updated when making changes within the content editable text box. 

without this it will rerender the schema on each change of these values, which moves the cursor to the end of the content editable (as well as being janky).

hand-dot:
@peteward 
Do I need to consider Forms or Viewers for this AdvancedTextSchema? 

Understanding the specific user flow in which it will be used would help in creating something even better.

packages/schemas/src/advancedText/propPanel.ts
```diff
@@ -0,0 +1 @@
+// TODO:
```
hand-dot, line 1: @peteward I can't build due to this file.

```
src/advancedText/index.ts:3:27 - error TS2306: File '/Users/user/Develop/pdfme/packages/schemas/src/advancedText/propPanel.ts' is not a module.

3 import { propPanel } from './propPanel.js';
                            ~~~~~~~~~~~~~~~~

src/advancedText/propPanel.ts:1:1 - error TS1208: 'propPanel.ts' cannot be compiled under '--isolatedModules' because it is considered a global script file. Add an import, export, or an empty 'export {}' statement to make it a module.

1 // TODO:



Found 2 errors in 2 files.
```

packages/schemas/src/advancedText/uiRender.ts
```diff
@@ -0,0 +1,45 @@
...
+    } = arg;
+
+    let dynamicVariablesStr = JSON.stringify(getUniqueVariableNames(value));
+
+    const renderArgs = {
+        value: schema.content || '',
+        schema,
+        rootElement,
+        onChange: (value: string) => {
+            if (onChange && onCustomAttributeChange) onCustomAttributeChange('content', value);
```
hand-dot, line 23: Ah... I got it.
We can change schema prop by onCustomAttributeChange.
Cool.

packages/schemas/src/advancedText/propPanel.ts
peteward: It won't build yet, it's not finished, I was just sharing the code

packages/common/src/types.ts
```diff
@@ -59,6 +59,7 @@ export interface PDFRenderProps<T extends Schema> {
  * @property {string} key - The key of the schema object.
  * @property {string} value - The string used for UI rendering.
  * @property {(value: string) => void} [onChange] - Used to change the value. Only applicable when the mode is 'form' or 'designer'.
+ * @property {(key: string, value: string) => void} [onChange] - Used to change another attribute for the schema. Only applicable when the mode is 'form' or 'designer'.
```
hand-dot, line 62: onChange → onCustomAttributeChange

 * @property {(key: string, value: string) => void} [onCustomAttributeChange] - Used to change another attribute for the schema. Only applicable when the mode is 'form' or 'designer'.

packages/schemas/src/advancedText/propPanel.ts
hand-dot: Okay

hand-dot:
@peteward 

Isn't what you really want the ability to replace placeholders in the input when passing it to the ReadOnlyText and generate function? 

How about adding the ability to customize pre-processing and post-processing for the generate function?

```ts
const inputs = [
  { headline: { name: "han-dot", favoriteColor: "blue" }, address: "aa-aa-a" },
];

const template = {
  schemas: [
    {
      headline: {
        type: "text",
        position: { x: 4.69, y: 7.15 },
        width: 186.51,
        height: 41.98,
        rotate: 0,
        opacity: 1,
        fontSize: 36,
        fontColor: "#14b351",
        fontName: "NotoSerifJP-Regular",
        readOnly: true,
        readOnlyValue: "Hi {name}, Your favorite color is {favoriteColor}",
      },
      address: {
        type: "text",
        position: { x: 4.69, y: 7.15 },
        width: 186.51,
        height: 41.98,
        rotate: 0,
        opacity: 1,
        fontSize: 36,
        fontColor: "#14b351",
        fontName: "NotoSerifJP-Regular",
      },
    },
  ],
  basePdf: "base64pdf...",
};

const pdf = await generate({
  inputs,
  template,
  options: {
    preprocessing: ({ inputs, template }) => {
      const newInputs = inputs.map((input) => {
        const newInput = {};
        Object.keys(input).forEach((key) => {
          const value = input[key];
          if (typeof value === "object") {
            const schemaIndex = template.schemas.findIndex(
              (schema) => schema[key]
            );
            const schema = template.schemas[schemaIndex][key];
            const { readOnlyValue } = schema;
            newInput[key] = readOnlyValue.replace(
              /{([^}]+)}/g,
              (_, key) => value[key]
            );
          } else {
            newInput[key] = key;
          }
        });
        return newInput;
      });
      return { inputs: newInputs, template };
    },
  },
});
```

This code is just an idea. I'm not sure if it will work as is, but by incorporating this process into the following part of the generate function, 

I think it would be possible to replace placeholders with inputs when creating the PDF.
https://github.com/pdfme/pdfme/blob/700a08852755a20691fc63e67e3b998aa14b6abf/packages/generator/src/generate.ts#L21

peteward:
> @peteward Do I need to consider Forms or Viewers for this AdvancedTextSchema?
> 
> Understanding the specific user flow in which it will be used would help in creating something even better.

I did not intend on implementing Form or Viewer support as we only use the designer.

If it was going to be more widely used then I would consider it.

peteward:
> @peteward
> 
> Isn't what you really want the ability to replace placeholders in the input when passing it to the ReadOnlyText and generate function?
> 
> How about adding the ability to customize pre-processing and post-processing for the generate function?
> 
> ```ts
> const inputs = [
>   { headline: { name: "han-dot", favoriteColor: "blue" }, address: "aa-aa-a" },
> ];
> 
> const template = {
>   schemas: [
>     {
>       headline: {
>         type: "text",
>         position: { x: 4.69, y: 7.15 },
>         width: 186.51,
>         height: 41.98,
>         rotate: 0,
>         opacity: 1,
>         fontSize: 36,
>         fontColor: "#14b351",
>         fontName: "NotoSerifJP-Regular",
>         readOnly: true,
>         readOnlyValue: "Hi {name}, Your favorite color is {favoriteColor}",
>       },
>       address: {
>         type: "text",
>         position: { x: 4.69, y: 7.15 },
>         width: 186.51,
>         height: 41.98,
>         rotate: 0,
>         opacity: 1,
>         fontSize: 36,
>         fontColor: "#14b351",
>         fontName: "NotoSerifJP-Regular",
>       },
>     },
>   ],
>   basePdf: "base64pdf...",
> };
> 
> const pdf = await generate({
>   inputs,
>   template,
>   options: {
>     preprocessing: ({ inputs, template }) => {
>       const newInputs = inputs.map((input) => {
>         const newInput = {};
>         Object.keys(input).forEach((key) => {
>           const value = input[key];
>           if (typeof value === "object") {
>             const schemaIndex = template.schemas.findIndex(
>               (schema) => schema[key]
>             );
>             const schema = template.schemas[schemaIndex][key];
>             const { readOnlyValue } = schema;
>             newInput[key] = readOnlyValue.replace(
>               /{([^}]+)}/g,
>               (_, key) => value[key]
>             );
>           } else {
>             newInput[key] = key;
>           }
>         });
>         return newInput;
>       });
>       return { inputs: newInputs, template };
>     },
>   },
> });
> ```
> 
> This code is just an idea. I'm not sure if it will work as is, but by incorporating this process into the following part of the generate function,
> 
> I think it would be possible to replace placeholders with inputs when creating the PDF.
> 
> https://github.com/pdfme/pdfme/blob/700a08852755a20691fc63e67e3b998aa14b6abf/packages/generator/src/generate.ts#L21

Yes, pre-processing would be a slightly cleaner way of doing it. But otherwise it is mostly already possible.

I can certainly use `readOnlyValue`. It didn't exist when I started writing the plugin so I used `content`, but it's just a label I guess. `content` feels more versatile for other plugin developers, but it doesn't really matter.

I think what's really cool is that most of what I am trying to achieve is already possible with the plugin system, just a few minor tweaks needed 😄 

I will continue to refine this over the next few days then share the result.

hand-dot:
Your advice has been quite helpful. Indeed, "content" is more versatile than "readOnlyValue".

There's a schema named SchemaForUIAdditionalInfo in `packages/common/src/schema.ts`, which has a property named `data`.

I'm thinking of renaming the `readOnlyValue` and `data` property names to `content` to reduce the code while maintaining functionality. I'll give it a try.

Can you keep this PR open until the desired functionality is achieved? I believe the plugin you're developing will be useful for others as well.

peteward:
Of course, thank you!

hand-dot:
ref: https://support.pdfgeneratorapi.com/en/category/expression-language-q203pa/

peteward:
Hi @hand-dot,

Most of the functionality that I intended for this plugin is done and working now. It needs some tidying and tests still but the result looks good:

https://github.com/pdfme/pdfme/assets/7068515/84d5164e-da7b-47e6-be99-15006aba58bd

packages/schemas/src/multiVariableText/uiRender.ts
```diff
@@ -0,0 +1,85 @@
+import { UIRenderProps } from '@pdfme/common';
+import { MultiVariableTextSchema } from './types';
+import { uiRender as parentUiRender } from '../text/uiRender';
+import { isEditable } from '../utils';
+import { substituteVariables } from './helper';
+
+export const uiRender = async (arg: UIRenderProps<MultiVariableTextSchema>) => {
+  const { value, schema, rootElement, mode, onChange, onSchemaAttributeChange, ...rest } = arg;
+
+  // This plugin currently does not support editing in form view, setting as read-only.
```
hand-dot, line 10: 👀

hand-dot:
I'm glad that I was able to create what you intended.

If there are any other necessary implementations, please let me know.

peteward:
Hi @hand-dot,

We are currently using a fork of pdfme because we have an old custom text schema similar to what this plugin achieves, although inferior. If my company can start using this plugin then we can stop using the fork and start updating to new versions more easily.

To do this we would either need:

1. Accept this plugin into core pdfme (or use it as the inspiration for a similar new text plugin). 
2. Allow me to make a PR with the 3 changes to pdfme that allow this plugin to work:
  * `onSchemaAttributeChange`
  * add an `id` attribute to the rendered text schema  
  * optimised `schemaRendererState` so that input focus is not lost when the schema changes  

Let me know what you think,
Thanks!

hand-dot:
@peteward 

In v4, it is possible to achieve the same functionality as `onSchemaAttributeChange`, so instead of creating a new PR, it might be okay to implement it on the v4 branch.

Reference: https://pdfme.com/blog/developing-pdfme-v4#about-backward-compatibility

peteward:
ahh ok, I will need to rework it then 👍

Issue #332, opened 2023-11-25
Add DynamicTable Schema


hand-dot:
# UPDATE
- [Developing pdfme V4](https://pdfme.com/blog/developing-pdfme-v4) - 2024/03/02
---

- PR: https://github.com/pdfme/pdfme/pull/408
- Preview: https://pdfme-playground-v4.vercel.app/

---

## Known Bugs

🔥 = priority

- [ ] #428
- [ ] #429
- [ ] #430
- [ ] #433
- [x] #431
- [x] #432

---

This issue is blocked by the following issue:
- #320 
- #272 
- #111

---

ref: 
- The thread that was discussed until this issue was created: https://github.com/pdfme/pdfme/discussions/148
- A repository implementing table functionality using pdf-lib: https://github.com/MP70/pdf-lib-draw-table
- Ideas on how to implement dynamic tables in pdfme currently: https://pdfme.com/blog/implementation-Ideas-for-dynamic-tables
- [Archive] Table schema implementation status: https://hand-dot.notion.site/Implementation-of-a-table-ade0377a8c06485dbe07e5b94bad68c8

MP70:
@hand-dot Can I ask how you're planning to draw the table on the pdf once you have added this to the schema?

hand-dot:
@MP70 

Thank you.
Firstly, regarding the support for links, I hadn't planned to include it in V4, but if it's necessary, I can add it.
https://github.com/pdfme/pdfme/issues/319

The idea of performing drawing via callbacks is very interesting. This means that the table schema would be limited to calculating the layout and style of the table, and the content of the table would be drawn through callbacks, right?

As for measuring the height and width of text, I'm not yet sure which method to use, but generally, the text schema of pdfme, including the dynamic font size feature, works well. It functions correctly not only with Latin characters but also with Japanese and Chinese fonts. However, it's not perfect. There was a recent bug report titled 'Line breaks in RTL languages'.

What I want to clarify is what features this DynamicTable Schema will have. I want to define the requirements before writing any code. I also want to make sure that these features are not overly complex.

Pdfme aims to be a user-friendly PDF library. Therefore, I think it's necessary to discard overly specialized features. The code you've written was originally developed for in-house use and then extracted. Before integrating it into the core library of pdfme, I want to clarify what is necessary and what is not.


Is there a way to check what features the current pdf-lib-draw-table supports and what requirements it had?

MP70:
[...]

Issue #111, opened 2022-11-22
Support adding new pages to template


asanoviskhak:
Hi there! I love what you are doing. I was playing around your library and found out that you can't add new page dynamically in editing modes. It would be great to have that, thanks!

asanoviskhak:
Mainly it would be creating PDF from blank, and not importing BASE_PDF, or build on top of BASE_PDF

hand-dot:
Hi! Thank you for the feature suggestion👍

Could you please tell me more detail and user interface and function interface?

hand-dot:
@asanoviskhak 

I labeled the answer because I did not receive a reply.
If there are no replies at this point, it will be closed after 30 days.
If you have any questions, please comment.

halindraprakoso:
I'm probably has the same idea. Basically, afaik there's no option to start from blank in the template designer on the website as users are currently required to make and use a blank pdf as the basePDF

xips1337:
Поддерживаю идею!

ujjwal-neu:
any updates on this feature or a way around to achieve this?

sushi2all:
The idea could simply be to have a "Add a blank page" section, with a height and width definition and then regenerate the template base pdf.

See relevant `pdf-lib` [docs](https://pdf-lib.js.org/docs/api/classes/pdfdocument#addpage)

Ben-CA:
Would also like to see multi-page support.

peteward:
> Would also like to see multi-page support.

pdfme is does support multiple pages, just not dynamically creating them for an existing base template/pdf.

This shouldn't be too hard to implement if someone is willing to contribute.

urenpatelce:
@peteward as you mentioned that pdfme supports multiple pages, so then what will be the template schema? is it possible to have the multiple pages designing at the same time ?

hand-dot:
implemented by https://github.com/pdfme/pdfme/pull/394

mohameddev19:
> implemented by [#394](https://github.com/pdfme/pdfme/pull/394)

First and foremost, I wish to commend your tremendous efforts and extend my deepest gratitude for them, and I truly regret that I was unable to assist.
Following the library installation, the 'Add New Page' button is absent from the Designer, although it is visible and functional in the Playground.

PR #389, opened 2023-12-23
Reconsider the internal data types and the template types

hand-dot:
Use content instead of some similar property.
ref: https://github.com/pdfme/pdfme/pull/388#issuecomment-1868260918
- [x] Change SchemaForUI's data and readOnlyValue to content

---

Refactor Template type
ref: https://github.com/pdfme/pdfme/pull/389#discussion_r1435775750
- [x] Remove `sampledata` and instead always use `content`.
- [x] Use `content` regardless of `readOnly` or not.
- [x] Eliminate `defaultValue` and replace it with always using `content`.
- [x] If possible, remove `columns`. Previously, `columns` played a role in the UI package for the zIndex and tabIndex of each schema. In the generator package, it was used to determine the rendering order of each schema, but this can be substituted with the order of keys in each schema.

---

It's wonderful that the data definition of templates can be simplified. However, if this gets merged, it will necessitate a major version upgrade, and users will need to migrate the templates they have created so far. 

I am positive about this change, but it might be better to complete the remaining two tasks that need to be implemented in V4 and then release it all together in V4.
- #111
- #332


packages/ui/src/helper.ts
```diff
@@ -361,8 +360,10 @@ export const fmtTemplate = (template: Template, schemasList: SchemaForUI[][]): T
         delete cur.id;
         // @ts-ignore
         delete cur.key;
-        // @ts-ignore
-        delete cur.data;
+        if (!cur.readOnly) {
+          // @ts-ignore
+          delete cur.content;
+        }
```
peteward, line 366: hmm so you can only use `content` if it's readOnly? 🤔 

I'm not sure how this will work for the use-case of content added by the editor and inputs, such as my case.

I was intending to use `data` for the JSON array of variables and `content` the editors text. 

I will think about it more tomorrow.

hand-dot: Ah, indeed. We could consider removing `sampledata` and instead always use `content`. This way, `content` can be used whether it's `readOnly` or not. It's also feasible to set `propPanel`'s `defaultValue` in the schema to use `content`.

Of course, since `sampledata` will be eliminated, this would be a breaking change, but I've been thinking of eventually removing `sampledata` and `columns`.

Regardless of whether we actually implement it, there's potential for further simplification, so I'll try modifying it to use `content` instead of `sampledata`.

peteward: OK. 

How would that work for my case where there's both `content` and `inputs` at the same time? 

content-inputs-data

Currently, `schema.data` models the inputs but you have consolidated these together above.

hand-dot: In your case, input/data is actually a property of the schema. It cannot be substituted for the inputs passed to the generator.

In my understanding, placeholders are automatically replaced by the customer's data, but how exactly is the `Input/Data` in the `propPanel` configured?

peteward: For the `propPanel` I was going to set the sample values added on `schema.data`. 
These would then be available as `sampledata` which could then  used for substitution on both the editor UI when not in editing mode, and when previewing the PDF.

I guess my issue is that `content` != `input`, i need both to be separate, and I think others writing plugins might need this in future. 

I think there are 3 cases:

* `content` == `input`  (pdfme original behaviour)
* `content` with no input (ready only version)
* `content` && `input` (a combination of both together)

I think any redesign should consider all of these cases

hand-dot: Thank you for the clear explanation.

To keep it simple while addressing the three cases, I think a custom preprocessing of the generator function is necessary. https://github.com/pdfme/pdfme/pull/388#issuecomment-1868248936

Don't you think this approach is good?

peteward:
Yes, preprocessor is a great idea.

You can write a preprocessor which will transform the content to include the values provided by input - this is no problem when input is provided.

However, if you remove `sampledata`, how will the generate preview function work in scenarios 1 & 3?  You either need to extract the content values and pass them back in as input for generation, or you don't pass anything and the schemas (or the generator) need to know how to work either without input data, or in a specific preview mode.

In my case, I wouldn't want the omission of an input value to be confused with rendering sample data, I would prefer rendering to be explicit values.

I agree that keeping `sampledata` as well as `content` in some cases feels like duplication, but it's either handling complexity or duplication.

🤔

hand-dot:
> However, if you remove `sampledata`, how will the generate preview function work in scenarios 1 & 3? 
- For scenario 1, the preview is generated based on the inputs from the Form and Viewer. However, in read-only cases, it uses the `content` from the schema. Reference: https://github.com/pdfme/pdfme/blob/383b986ea88f52e13a09bf9b5daef6286d3d72ba/packages/ui/src/components/Preview.tsx#L98
- For scenario 3, the preview is generated based on the input. If you want to use the `content` from the schema, you need to transform the input placeholders when doing Viewer.setInputs. In the case of the Form preview, the placeholders will be displayed as they are, which I think is a natural behavior.

hand-dot:
For now, I've merged it into the v4 branch. This fix, along with other breaking changes, will be implemented for other [v4 projects](https://github.com/orgs/pdfme/projects/8) on the v4 branch.

For the time being, it won't be merged into the main branch, so feel free to share any opinions in the discussion thread below.
https://github.com/pdfme/pdfme/discussions/393

peteward:
This is awesome.

Thank you so much @hand-dot.

When I have a working version of my plugin I will rebase onto these changes and try them out 👍

PR #390, opened 2023-12-25
Performance test handling

dkastl:
This pull request introduces different error handling of performance related tests for CI and non-CI.

When environment variable `CI` is set to `true` (i.e. in GH Actions `test` workflow), the test fails with an error in case the execution time constraints are violated. 
In other cases a constraint violation triggers a warning. 

The `PERFORMANCE_THRESHOLD` (in seconds) can be also set as environment variable.

Reference discussion: https://github.com/pdfme/pdfme/discussions/387#discussioncomment-7940652

### DIFF
--- a/packages/generator/__tests__/integration1.test.ts
+++ b/packages/generator/__tests__/integration1.test.ts
@@ -4,6 +4,8 @@ import { label, envelope } from './assets/templates';
 import { text, image } from '@pdfme/schemas';
 import { getFont, getPdf, getPdfTmpPath, getPdfAssertPath } from './utils';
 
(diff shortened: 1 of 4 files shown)


hand-dot:
@dkastl 
Aren't there any modifications to .github/workflows/test.yml?

dkastl:
No, it should not be necessary. The `CI` environment variable should be already set, see https://github.com/pdfme/pdfme/blob/main/.github/workflows/test.yml#L23

hand-dot:
Thank you @dkastl !!

peteward:
very nice @dkastl 👏

Issue #392, opened 2023-12-27
New basePdf type


hand-dot:
Currently, it's possible to pass a base64 string, Uint8Array, or ArrayBuffer, but I intend to enable specifying a blank PDF by setting parameters like width, height, and margin.

---

ref: https://github.com/pdfme/pdfme/issues/111#issuecomment-1717526476

hand-dot:
implemented by https://github.com/pdfme/pdfme/pull/394

PR #394, opened 2023-12-27
New basePdf type and Support adding new pages to template

hand-dot:
ref:
- #111 
- #392 

---

TODO
- [x] padding (This is used for page breaks by dynamic tables and the draggable area in the designer.)
- [x] add / remove page from designer function
- [x] add/remove page button ui design
- [ ] ~~add ui to set padding~~
- [ ] ~~add demo to website and playground(because this feature is not available if basePdf doesn't have width and height this feature will not be shown existing site)~~ -> #401

---

preview

https://github.com/pdfme/pdfme/assets/24843808/000501af-76e8-43de-a7e0-b56f38519c3c


packages/generator/src/helper.ts
```diff
@@ -1,68 +1,76 @@
...
+  Plugins,
+  GeneratorOptions,
+  Template,
+  PDFRenderProps,
+  Plugin,
+} from '@pdfme/common';
 import { builtInPlugins } from '@pdfme/schemas';
 import { PDFPage, PDFDocument, PDFEmbeddedPage, TransformationMatrix } from '@pdfme/pdf-lib';
-import { getB64BasePdf, BasePdf } from '@pdfme/common';
+import { getB64BasePdf, BasePdf, isBlankPdf, mm2pt } from '@pdfme/common';
```
peteward, line 12: this can be combined with the import above

hand-dot: Oh, you are right

hand-dot:
@peteward 

I'm unsure where to add the buttons. What do you think about this arrangement?

CleanShot 2023-12-28 at 11 31 11@2x

hand-dot:
Update 
CleanShot 2023-12-28 at 18 15 01@2x

hand-dot:
https://github.com/pdfme/pdfme/assets/24843808/b0b5f915-8ffe-4378-9d6d-a2e93ca82088

packages/generator/__tests__/assets/templates/pet.json
```diff
@@ -404,7 +404,16 @@
...
+  "basePdf": {
```
hand-dot, line 407: The main part of what I did in this PR is that I made it so that we can set it with a new type called [BlankPdf](https://github.com/pdfme/pdfme/pull/394/files#diff-137a58655e18fd31de35b0116e6e0e215889557a42c396f63f0479c20f3bba9fR87) instead of PDF data in basePdf.

hand-dot (commented):
@peteward 
If you have time, could you check this PR? 
Of course, I would prefer you to prioritize family time over the review.

packages/ui/src/components/Designer/Sidebar/DetailView/index.tsx
```diff
@@ -123,8 +123,8 @@ Check this document: https://pdfme.com/docs/custom-schemas`);
       key: { title: i18n('fieldName'), type: 'string', required: true, span: 14 },
       '-': { type: 'void', widget: 'Divider' },
       align: { title: i18n('align'), type: 'void', widget: 'AlignWidget' },
-      x: { title: 'X', type: 'number', widget: 'inputNumber', required: true, span: 8 },
-      y: { title: 'Y', type: 'number', widget: 'inputNumber', required: true, span: 8 },
+      x: { title: 'X', type: 'number', widget: 'inputNumber', required: true, span: 8, min: 0 },
```
peteward, line 126: I think we might want to consider schema that start outside of the window. Although I haven't tried this myself so I don't know how it would render 🤔  But I could imagine both text where you want to rotate and position tight to the left edge (negative X pos) or similarly for images where you only want part of the image to be in canvas.

As I say, I haven't tried either of these so I don't know how it would render.

packages/ui/src/components/Designer/index.tsx
```diff
@@ -112,13 +120,29 @@ const TemplateEditor = ({
...
+            tgt.position.x = Math.min(
+              Math.max(Number(value), paddingLeft),
+              pageWidth - targetWidth - paddingRight
+            );
+          } else {
+            tgt.position.y = Math.min(
+              Math.max(Number(value), paddingTop),
+              pageHeight - targetHeight - paddingBottom
+            );
+          }
```
peteward, line 138: it might be worth extracting this out to a testable function, this is within a really important callback function

packages/ui/src/components/Designer/index.tsx
```diff
@@ -206,15 +262,17 @@ Check this document: https://pdfme.com/docs/custom-schemas`);
         }}
         zoomLevel={zoomLevel}
         setZoomLevel={setZoomLevel}
+        addPageAfter={isBlankPdf(template.basePdf) ? handleAddPageAfter : undefined}
+        removePage={isBlankPdf(template.basePdf) ? handleRemovePage : undefined}
```
peteward, line 266: is `isBlankPdf` efficient? should we store the result to avoid the additional call?

peteward (commented):
great, some minor comments

packages/ui/src/components/Designer/index.tsx
hand-dot: You're right. This function is becoming complex. I have created a separate issue.

https://github.com/pdfme/pdfme/issues/402

hand-dot: ref: https://github.com/pdfme/pdfme/pull/403

vpn-dev:
@hand-dot  i am not seeing any option to add pages while in designer, can you help me with it

PR #395, opened 2023-12-27
promote includetext from  bwipjs arg that is always set to true to schema option (same as textColor)

### LINKED ISSUE #23: Add include text option to  in the barcode schema
ref: https://github.com/hand-dot/labelmake/issues/33

rhurling:
I'm pretty sure I implemented the bwipjs arg includetext at all the necessary spots, the only missing thing should be the new label in i18n.
This PR is also currently untested, I should be able to test it tomorrow.

fixes #23

### DIFF
--- a/packages/schemas/src/barcodes/helper.ts
+++ b/packages/schemas/src/barcodes/helper.ts
@@ -1,7 +1,7 @@
 import { b64toUint8Array } from '@pdfme/common';
 import bwipjs, { RenderOptions } from 'bwip-js';
 import { Buffer } from 'buffer';
-import { BARCODE_TYPES } from './constants.js';
+import { BARCODE_TYPES, DEFAULT_BARCODE_INCLUDETEXT } from './constants.js';
 import { BarcodeTypes } from './types';
 
 // GTIN-13, GTIN-8, GTIN-12, GTIN-14
@@ -143,7 +143,7 @@ export const createBarCode = async (arg: {
     backgroundColor,
     barColor,
     textColor,
-    includetext = true,
+    includetext = DEFAULT_BARCODE_INCLUDETEXT,
   } = arg;
 
   const bcid = barCodeType2Bcid(type);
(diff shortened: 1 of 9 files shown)


packages/ui/src/i18n.ts
hand-dot, line 1: Can you set 'includetext' on all dict?
Otherwise the build will fail.

hand-dot (commented):
Hey @rhurling!!

Awesome. I woke up feeling great this morning when I saw this PR. Thank you.

I'd like you to fix the build since it's failing.
After that, I'll do a functionality check on my end.

hand-dot:
@rhurling Thanks
I gonna check this PR tomorrow

packages/schemas/src/barcodes/propPanel.ts
```diff
@@ -181,7 +196,14 @@ export const getPropPanelByBarcodeType = (barcodeType: string): PropPanel<Barcod
         ],
       },
       ...(barcodeHasText
-        ? { textColor: { title: i18n('schemas.textColor'), type: 'string', widget: 'color' } }
+        ? {
+            textColor: { title: i18n('schemas.textColor'), type: 'string', widget: 'color' },
+            includetext: {
+              title: i18n('schemas.barcodes.includetext'),
+              type: 'boolean',
+              widget: 'switch',
```
hand-dot, line 204: I changed it to a switch when placed next to a color input because I preferred the look. It's just a matter of personal preference.

CleanShot 2023-12-29 at 13 21 30@2x

↓

CleanShot 2023-12-29 at 13 21 01@2x

packages/schemas/src/barcodes/helper.ts
```diff
@@ -133,11 +133,20 @@ export const createBarCode = async (arg: {
...
-  const { type, input, width, height, backgroundColor, barColor, textColor } = arg;
+  const {
+    type,
+    input,
+    width,
+    height,
+    backgroundColor,
+    barColor,
+    textColor,
+    includetext = DEFAULT_BARCODE_INCLUDETEXT,
```
hand-dot, line 146: We need to insert a default value here, as it would change the behavior of existing templates.

hand-dot (approved):

hand-dot:
@rhurling 
I made a few minor adjustments, but basically, there were no issues. Awesome!

Issue #397, opened 2023-12-28
Add a version number to pdfme template from V4 onwards


peteward:
### Is your feature request related to a problem? Please describe.

V4 will likely introduce significant BC breaks. We need a way to determine what version a template file is to help user applications manage these differences

### Describe the solution you'd like

Add a version number attribute to the template file to indicate which version of pdfme this was built with.

It could either be the version of pdfme itself, or a version of the template file format, as this should change much less often. I would suggest starting with version 2 in this case.

### Describe alternatives you've considered

-

hand-dot:
There's no doubt that we need to manage the version of the template. 

Basically, I want to align major version upgrades of pdfme with the timing when template backward compatibility is lost. 
So, I think it's better to match it with the major version of pdfme itself.

PR #404, opened 2024-01-04
Add a version number to pdfme template from V4 onwards

hand-dot:
ref: Add a version number to pdfme template from V4 onwards #397


peteward (approved):
great

PR #417, opened 2024-01-31
Fix pdf text render overlap by avoiding line breaks in pdflib

### LINKED ISSUE #398: Line breaks in RTL languages (Arabic in this case)
### Describe the bug

Hi everyone,
Having a long string (with no line breaks)  results in weird line breaks :
image

this is the design template 
image

And these are the parameters
image

i tried to changes petameters (alignment, font size ..) but it still resulted in those line breaks. I temporarily fixed the issue by counting how many Arabic characters can fit in that rectangle without line break, and then i manually add line breaks :
`
function getDescription(text?: string): string {
	const words = (text?.split(/\s+/) ?? []).filter((c) => !['\n', '\r'].includes(c));
        // number of Arabic characters (approximately) that can fit without resulting in line break 
	const maxCharacters = 80;
	let charCounter = 0;
	let result = '';

	for (const word of words) {
		charCounter += word.length;

		if (charCounter <= maxCharacters) {
			result += word + ' ';
		} else {
			result += '\n' + word + ' ';
			charCounter = word.length;
		}
	}

	return result.trim();
}
`
caveman solution but for now it seems to work
image

I was wondering if there's a solution for this, or am i just doing something wrong.
Have a great day everyone 🙏

### To Reproduce

Just use a font that support arabic characters and a long arabic text with no line breaks

### Expected behavior

for line breaks to properly work 

### Your Environment

```markdown
- pdfme version:
- Operating system:
```

### Your Error Log

```shell
-
```

### Additional context

_No response_

peteward:
fixes #398 

Pdfme calculates how lines should be split and renders them individually. It does this for a number of reasons including alignment, dynamic font size calculation and rotation. 

Pdflib has it's own line breaking which, in theory, shouldn't happen if calculations between the 2 libraries match up. However, pdflib appears to be calculating the width of a line of some fonts differently (incorrectly) compared to pdfme. Regardless of what and why this difference is (it wasn't immediately obvious to me), we don't need pdflib to split lines because we've already done it.

You cannot tell pdflib not to split lines, but we can make it impossible through this change.

pdflib `drawText()` function does the following:

```
    const lines =
      options.maxWidth === undefined
        ? lineSplit(cleanText(text))
        : breakTextIntoLines(text, wordBreaks, options.maxWidth, textWidth);
```

Previously we pass the `maxWidth` option which will result in `breakTextIntoLines()` being called. This is a more intelligent line breaking algorithm which is very similar to what pdfme does (although comes up with a different result).

By not passing maxWidth, we instead will call `lineSplit()`, which is simply the following:

```
export const lineSplit = (text: string) => text.split(/[\n\f\r\u000B]/);
```

By updating our split regex to include `\f` (form feed) and `u000B` ([line tabulation](https://codepoints.net/U+000B?lang=en)) we can ensure that no splitting will happen in pdf rendering, even if those characters are unlikely in the input.

BEFORE with text overlapping using Cairo font:
Screenshot 2024-01-31 at 07 05 26

AFTER with no overlap:
Screenshot 2024-01-31 at 07 05 48

### DIFF
--- a/packages/schemas/src/text/pdfRender.ts
+++ b/packages/schemas/src/text/pdfRender.ts
@@ -126,7 +126,7 @@ export const pdfRender = async (arg: PDFRenderProps<TextSchema>) => {
   };
 
   let lines: string[] = [];
-  value.split(/\r\n|\r|\n/g).forEach((line: string) => {
+  value.split(/\r\n|\r|\n|\f|\u000B/g).forEach((line: string) => {
     lines = lines.concat(getSplittedLines(line, fontWidthCalcValues));
   });
 
@@ -175,9 +175,7 @@ export const pdfRender = async (arg: PDFRenderProps<TextSchema>) => {
       size: fontSize,
       color,
       lineHeight: lineHeight * fontSize,
-      maxWidth: width,
       font: pdfFontValue,
-      wordBreaks: [''],
       opacity,
     });
   });


hand-dot:
@peteward 
Great.
By adding |\f|\u000B, we can now correctly split patterns that were not covered before.

I plan to merge and release this later.

hand-dot (approved):

PR #436, opened 2024-02-29
Added support for color type option in generator

necessarylion:
```ts
const pdf = await generate({
    ...
    options: {
      ...
      font, 
      // now support CMYK or RGB color type
      // if the value is not set, hex color will be generated into RGB color type
      colorType: 'cmyk'
    },
});
```

Color type will applied to `text`, `line`, `rect` and `ellipse` which are the the components that need colors at the moment.


hand-dot:
Hey @necessarylion !!!
So cool! 

This weekend I will check this.

packages/schemas/src/utils.ts
```diff
@@ -103,6 +103,50 @@ export const hex2RgbColor = (hexString: string | undefined) => {
...
+
+    if (!isValid) {
+      throw new Error(`Invalid hex color value ${hexString}`);
+    }
+
+    // Remove the # if it's present
+    hexString = hexString.replace('#', '');
+
+    // Extract the hexadecimal color code and the opacity
+    var hexColor = hexString.substring(0, 6);
```
hand-dot, line 118: Please use `const` or `let` instead of `var`

hand-dot (changes requested):
Thank you for the great work. Please fix the part where `var` is used, as we would prefer not to use it.

packages/schemas/src/utils.ts
necessarylion: Thanks. I have updated.

hand-dot (approved):
LGTM

hand-dot:
@necessarylion 

Released! https://github.com/pdfme/pdfme/releases/tag/3.3.0

Thank you for your contribution 👍

PR #444, opened 2024-03-13
contentEditable firefox fix

valushagrinchik:
As `contentEditable = 'plaintext-only'` doesn't work in firefox, this workaround fixes it and keeps 'plaintext-only' functionality (avoiding of styles be pasted)

### DIFF
--- a/packages/schemas/src/text/helper.ts
+++ b/packages/schemas/src/text/helper.ts
@@ -318,3 +318,5 @@ export const calculateDynamicFontSize = async ({
 
   return dynamicFontSize;
 };
+
+export const isFirefox = () => navigator.userAgent.toLowerCase().indexOf('firefox') > -1;
--- a/packages/schemas/src/text/uiRender.ts
+++ b/packages/schemas/src/text/uiRender.ts
@@ -17,6 +17,7 @@ import {
   calculateDynamicFontSize,
   getFontKitFont,
   getBrowserVerticalFontAdjustments,
+  isFirefox,
 } from './helper.js';
 import { addAlphaToHex, isEditable } from '../utils.js';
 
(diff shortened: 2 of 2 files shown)


peteward (commented):
Hi @valushagrinchik !

Thanks for the contribution and finding this issue.

Whilst pasting is a common use case that this addresses, it's also possible to format text in `contentEditable=true` elements in other ways, such as Cmd/Ctrl + B/I/U (amongst others).

There is quite a lot to catch and consider here (I've only just started thinking about it). 

It might be sensible to consider firefox support as a polyfill, so use `plaintext-only` but in the case of Firefox use `true` with additional event handlers as you have added here.

This would be a more graceful experience for the wider group of browsers and still have some support for Firefox with caveats. 

It does seem that [the Firefox team will eventually implement the support](https://bugzilla.mozilla.org/show_bug.cgi?id=1291467), but no idea on timelines...

hand-dot:
Hey @valushagrinchik 

Thank you creating PR!!

@peteward 

>  it's also possible to format text in contentEditable=true elements in other ways, such as Cmd/Ctrl + B/I/U (amongst others).

Indeed, I hadn't considered this case.

I'd like to hear your opinion. After your advice, @valushagrinchik modified the process to change it using try-catch, but I wonder if it would be better to switch the process with a function like the one below?

```javascript
const isFirefox = () => navigator.userAgent.toLowerCase().indexOf("firefox") > -1;
```

I think there's no problem with the rest of it. I'd be very happy either way, as it will improve the critical issue of being unable to edit text in Firefox.

peteward:
Being explicit about the browser seems like a good idea. We know that it's only FF out of current browsers that has the issue so it would be safe. It's also clearer in code why there are different approaches.

However, there's more to consider. 
When you type in `contenteditable=true` you will automatically get `<div>` and `<br>` tags inserted when you type. We need to find a way to handle this: 

Screenshot 2024-03-14 at 05 47 25

peteward:
[This fiddle](https://github.com/pdfme/pdfme/pull/444) appears to show a solution for handling the keypresses and replacing with newline characters.

peteward:
@valushagrinchik would you mind updating to include the check for Firefox explicitly and handling newlines without creating HTML elements?

valushagrinchik:
> [This fiddle](https://github.com/pdfme/pdfme/pull/444) appears to show a solution for handling the keypresses and replacing with newline characters.

@peteward It looks like incorrect link is attached, so that I couldn't check what you suggested

peteward:
Hi @valushagrinchik ,

apologies, here is the correct link: https://jsfiddle.net/1te5hwv0/

I think when you paste content in it might be OK, but when you press ENTER while typing it creates `<div>` elements for each line. 

The screenshot above was from your branch which I tried locally.

peteward:
Hi @valushagrinchik,

Are you working on this at the moment? If not I can try to finish it today.

Thanks.

peteward:
Hi @valushagrinchik,

Good job! This works well for me when testing locally. The dev tools confusingly show <br> elements but these don't appear to exist when you export the template or render to PDF, so guessing this is just a quirk with devtools and how it's rendering the new lines.

@hand-dot can you review please?

hand-dot:
Thank you @peteward 
I'll test it tomorrow morning and then release it!

hand-dot:
Released! https://github.com/pdfme/pdfme/releases/tag/3.4.2

Thank you @valushagrinchik and @peteward !

Issue #442, opened 2024-03-07
Make pageCursor publicly available in the Designer 


Letsmoe:
### Is your feature request related to a problem? Please describe.

When working with multiple pages it's impossible to tell which page we're currently on without inspecting the dom or writing your own page scroll observer making it unnecessarily difficult to use a schema for each page individually.

### Describe the solution you'd like

The `pageCursor` property already exists, it would just be a matter of exposing it as a public property of the `Designer` class.

### Describe alternatives you've considered

/

### Additional context

I'd be happy to create a pr for this.

hand-dot:
Hey @Letsmoe, Could you explain a bit more why we should add this feature?

Letsmoe:
If I'm not mistaken it's not currently possible to tell which page the user is on in the designer, right? That makes it practically impossible to put new fields into the correct schema objects and since there is no way to move fields between pages in the designer you basically cannot place objects on the pages following the first one. Is there maybe another way that I simply overlooked? 
Otherwise having the current page available would be a major thing especially when creating new pages on the fly and auto-generating page numbers for them.

hand-dot:
Thank you @Letsmoe ,

Does the following PR seem to solve your problem?
https://github.com/pdfme/pdfme/pull/445?

Letsmoe:
Yep, sure does!

hand-dot:
Perfect!!!

hand-dot:
This Request was handled by https://github.com/pdfme/pdfme/pull/446
We can use `getPageCursor` from pdfme@3.4.1.

PR #445, opened 2024-03-14
Exposed page cursor from Designer

hlin-neo4j:
Thought it might be helpful to be to add an api to the Designer to get the current page.

### DIFF
--- a/packages/ui/src/Designer.tsx
+++ b/packages/ui/src/Designer.tsx
@@ -16,6 +16,7 @@ import { cloneDeep } from './helper.js';
 class Designer extends BaseUIClass {
   private onSaveTemplateCallback?: (template: Template) => void;
   private onChangeTemplateCallback?: (template: Template) => void;
+  private pageCursor: number = 0;
 
   constructor(props: DesignerProps) {
     super(props);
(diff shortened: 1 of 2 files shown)


hand-dot:
Hey @hlin-neo4j ! Thank you for creating PR!

I've checked the code and behavior; it looks good. 
However, I don't quite understand the use case.

The same goes for the issue below.
https://github.com/pdfme/pdfme/issues/442

Could you explain a bit more why we should add this feature?

hlin-neo4j:
Ah I didn't see the #442 issue, but it sounds like there's at least two of us that needs it. For my use case, I have an external component (a sidebar) that displays and tracks the Schemas on the Designer. I store metadata about the Schemas as well, and all of this is stored in my application's Redux. I have users interact with this external Sidebar, so it is calling designer.current.updateTemplate, getTemplate, etc to make changes to the Designer. Now when the user creates multiple pages and scrolls between them, my Sidebar does not get any events that say a different page is displayed. designer.getTemplate is the only call that I know of that gives back information of the Template Schemas, but there is no "currentPage."

hand-dot:
@hlin-neo4j 
Thank you, I understand. 
This feature is very effective, so I want to release it soon. 

Could you switch the base branch to main? 🙏

hlin-neo4j:
@hand-dot I can pull back this pull request and make a new one off of base branch if that works for you

hand-dot:
Can you make new PR?🙇‍♂️

hlin-neo4j:
@hand-dot Just created #446

hand-dot:
Thank you, I'll handle in https://github.com/pdfme/pdfme/pull/446

Issue #400, opened 2023-12-30
Add a Left Sidebar for Placing Schemas


hand-dot:
### Is your feature request related to a problem? Please describe.

Currently, in pdfme, schemas are added by clicking the "Add new field" button. This button is located at the bottom of the right sidebar, and I don't think the user experience is particularly good. Not only is the location not optimal, but for example, if you want to add an image schema, you need to change the schema type after adding it.

### Describe the solution you'd like

Add a left sidebar, and place icons for each schema type on this sidebar. Allow schemas to be placed by dragging and dropping these icons.

![CleanShot 2023-12-30 at 15 32 54](https://github.com/pdfme/pdfme/assets/24843808/dd61bf1a-6740-4111-8293-a64cdb6dac68)

ref: https://pdfgeneratorapi.com/templates/free-templates/personalized-quote -> Edit template button

### Describe alternatives you've considered

None.

### Additional context

I have been thinking about this UX improvement for a while, but it required adding icons to existing schemas and would be a breaking change, so I haven't been able to implement it. However, with many breaking changes occurring in V4, I would like to take this opportunity to implement it.

peteward:
I really like this.

It also increases the vertical real-estate of the right-hand sidebar, which can scroll off the screen for smaller resolutions.

hand-dot:
ref: https://docs.dndkit.com/api-documentation/droppable

vpn-dev:
Any work in progress on this @hand-dot 
?

hand-dot:
Hi @vpn-dev 

here is pr: https://github.com/pdfme/pdfme/pull/452
implementation is done.

this feature will be release as v4

vpn-dev:
Cool @hand-dot , any idea when V4 will be published?

hand-dot:
@vpn-dev 
we need to resolve https://github.com/pdfme/pdfme/issues?q=is%3Aissue+is%3Aopen+label%3Atable

PR #452, opened 2024-03-22
Add a Left Sidebar for Placing Schemas #400

hand-dot:
ref: https://github.com/pdfme/pdfme/issues/400

CleanShot 2024-03-22 at 12 20 41


hand-dot:
Hi @peteward,
If you have time, Can you check this PR?

packages/ui/src/components/Designer/index.tsx
```diff
@@ -219,59 +226,85 @@ Check this document: https://pdfme.com/docs/custom-schemas`);
...
+        onDragEnd={(event) => {
+          // Triggered after a scheme is dragged & dropped from the left sidebar.
+          if (!event.active) return;
+          const active = event.active;
+
+          const rect = paperRefs.current[pageCursor].getBoundingClientRect();
+          const initialTop = (active.rect.current.initial?.top || 0) - rect.top;
+          const initialLeft = (active.rect.current.initial?.left || 0) - rect.left;
+          const _scale = scale < 1 ? scale + 1 : scale;
+          const adjust = 0.915; // TODO: Investigate later as to why it needs to be adjusted.
```
hand-dot, line 239: When dragging an element to the bottom of the page, its position gets misaligned.

Furthermore, if the page is zoomed in, the misalignment becomes even greater.

However, I have determined that this issue does not qualify as a release blocker, and we can fix it in the future.

packages/ui/src/components/Designer/index.tsx
```diff
@@ -27,6 +28,12 @@ import Root from '../Root';
 import ErrorScreen from '../ErrorScreen';
 import CtlBar from '../CtlBar';
 
+const px2mm = (px: number): number => {
```
peteward, line 31: we already have px2mm in the image schema, should we add it to the common/helper.ts file where the other conversion functions are?

packages/ui/src/components/Designer/index.tsx
```diff
@@ -219,59 +226,85 @@ Check this document: https://pdfme.com/docs/custom-schemas`);
...
-        activeElements={activeElements}
-        schemas={schemasList[pageCursor] ?? []}
-        changeSchemas={changeSchemas}
-        onSortEnd={onSortEnd}
-        onEdit={(id: string) => {
-          const editingElem = document.getElementById(id);
-          editingElem && onEdit([editingElem]);
+      <DndContext
+        onDragEnd={(event) => {
+          // Triggered after a scheme is dragged & dropped from the left sidebar.
```
peteward, line 231:
```suggestion
          // Triggered after a schema is dragged & dropped from the left sidebar.
```

playground/src/plugins/signature.ts
```diff
@@ -57,9 +57,9 @@ export const signature: Plugin<Signature> = {
   pdf: image.pdf,
   propPanel: {
     schema: {},
-
     defaultSchema: {
       type: 'signature',
+      // icon: '<svg xmlns="http://www.w3.org/2000/svg" width="24" height="24" viewBox="0 0 24 24" fill="none" stroke="currentColor" stroke-width="2" stroke-linecap="round" stroke-linejoin="round" class="lucide lucide-pen"><path d="M17 3a2.85 2.83 0 1 1 4 4L7.5 20.5 2 22l1.5-5.5Z"/></svg>',
```
peteward, line 62: do you intend to leave this commented?

peteward (commented):
This is very cool!

I would be happy to look at the positioning issues soon if you like?

playground/src/plugins/signature.ts
hand-dot: Yes, this is example if not set icon, left sidebar icon will be plugin's Key text

packages/ui/src/components/Designer/index.tsx
hand-dot: Ah, yes. It should be.

hand-dot:
Thank you @peteward !

> This is very cool!
> I would be happy to look at the positioning issues soon if you like?

Oh god. Please look this problem.

peteward:
Sure, let's get this merged into `v4` and I will try to help with the positioning. 👍

PR #454, opened 2024-03-24
Left sidebar schema drop placement fix

### LINKED ISSUE #453: Left sidebar drop positioning accuracy
### Describe the bug

When dragging in new schema from the left-hand sidebar they are not positioned correctly and the positioning inaccuracy increases the further away from the top-left corner

### To Reproduce

* build v4 branch
* Drag a schema from the left side bar
* Drop onto the canvas

### Expected behavior

The schema should be positioned exactly where it is dropped.

### Your Environment

```markdown
- pdfme version: v4 branch
```

### Your Error Log

```shell
-
```

### Additional context

The fix should consider:

* zooming
* multi-page PDFs
* bordered pdfs
* window size relative to pdf (zoomed in and unable to see full canvas)
* different schema types
* placing the editor in an area with both top and left sidebars outside of the canvas
* a combination of the above

peteward:
fixes #453 

This increases the accuracy of the drop placement.

There is one minor issue that I have compensated for, which is the canvas scale and how that impacts the initial drag position of the schema from it's icon. At a scale of `1` this is perfectly aligned with the top-left corner of the icon position. Any other scale shifts this position left or right based on how big the scale change is.

I have written a function to compensate for this. It is crude but it works quite well, although not 100% perfect. There is a tiny jump at some scales but I don't think this is a problem.

You can see below that when you first click on the icon to drag the schema it's starting position changes based on the scale:

https://github.com/pdfme/pdfme/assets/7068515/fae5daf5-d621-4e13-ba0b-75308f6c5470


hand-dot (approved):
LGTM!!

PR #463, opened 2024-03-29
Fix Spanish translations for v4

ikerd:
Seems like the latest Spanish translations added via merge break the build process, I have modified the Spanish translations to meet the v4 type and allow building successfully


hand-dot:
@ikerd Thank you!!!!

PR #466, opened 2024-03-31
Multi-var text plugin (V4)

peteward:
This plugin is functionally ready for review.  The behaviour is shown below:

https://github.com/pdfme/pdfme/assets/7068515/84d5164e-da7b-47e6-be99-15006aba58bd

The idea is to store the variables as a JSON string in `content` and the typed input as `text`.

This plugin will allow n variables to be added to a single text field, where n can be >= 0 (so this will also achieve a read-only schema). These variables are automatically detected as you type, and done so in an optimally efficient way that is usable even with dynamic font sizing enabled, as shown in the video above.

This plugin does not support the Form mode. Some thought would be needed as to the UI but it would be doable if people were interested. It would probably need an overlay.

### TODO:

- [ ] Add tests where applicable

### DIFF
--- a/packages/ui/src/components/Renderer.tsx
+++ b/packages/ui/src/components/Renderer.tsx
@@ -57,6 +57,10 @@ const Renderer = (props: RendererProps) => {
   const ref = useRef<HTMLDivElement>(null);
   const _cache = useRef<Map<any, any>>(new Map());
 
+  // If this schema is actively being edited (e.g. typing into a field)
+  // then we don't want changes to that schema triggering a re-render
+  const schemaRerenderState = mode === 'form' || mode === 'designer' ? '' : JSON.stringify(schema);
+
   useEffect(() => {
     if (ref.current && schema.type) {
       const render = Object.values(pluginsRegistry).find(
@@ -95,6 +99,7 @@ Check this document: https://pdfme.com/docs/custom-schemas`);
       }
     };
   }, [value, JSON.stringify(schema), JSON.stringify(options), mode, scale]);
+// }, [mode, scale, schemaRerenderState, JSON.stringify(options)]);
 
   return (
     <Wrapper {...props}>
--- a/packages/ui/src/components/Renderer.tsx
+++ b/packages/ui/src/components/Renderer.tsx
@@ -58,7 +58,7 @@ const Renderer = (props: RendererProps) => {
   const _cache = useRef<Map<any, any>>(new Map());
 
   // If this schema is actively being edited (e.g. typing into a field)
-  // then we don't want changes to that schema triggering a re-render
+  // then we don't want changes to that schema made elsewhere to trigger a re-render and lose focus.
   const schemaRerenderState = mode === 'form' || mode === 'designer' ? '' : JSON.stringify(schema);
 
   useEffect(() => {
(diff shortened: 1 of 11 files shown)


packages/ui/src/components/Renderer.tsx
```diff
@@ -94,7 +98,7 @@ Check this document: https://pdfme.com/docs/custom-schemas`);
         ref.current.innerHTML = '';
       }
     };
-  }, [value, JSON.stringify(schema), JSON.stringify(options), mode, scale]);
+}, [mode, scale, schemaRerenderState, JSON.stringify(options)]);
```
peteward, line 101: `value == schema.content` so changes would be caught anyway.

peteward: hmm this change doesn't work now for removing the image file on an image schema. This needs a re-think.

peteward:
Hi @hand-dot,

Just looking at the Form UI for this plugin.

There are 2 options. 

1). Add an overlay for filling in the variables that becomes visible when you select the schema, similar to the prop panel:

Screenshot 2024-04-11 at 08 33 47

This approach would be easier to implement and it would likely require minimal or no modifications to the existing text plugin. However, the UX isn't ideal.

2). Create content-editable sections for all of the variables within the schema:

https://github.com/pdfme/pdfme/assets/7068515/b6ee5412-262a-4c3e-8e7f-5880e1a2a089

This feels like better UX, but it will either require significant duplication of the existing text plugin or refactoring the existing text plugin into smaller functions so that aspects of it can be reused. There will be complexity here as typing within each field will need to consider dynamic font sizing of the overall text area. 

It is achievable, but let me know what you think and if you think I should continue the work.

hand-dot:
@peteward 

- I believe the ideas and use cases are strong and likely sought after by other users. Let's integrate these into the `schemas` package.
- For the implementation, I prefer the UX approach outlined in option 2.
  - To enhance user understanding during input, it would be better to have different designs for the outer border and the placeholder border.
  CleanShot 2024-04-13 at 16 50 36@2x
  
- Could you implement it according to option 2? I'd appreciate support until the release.

---

- I have another matter to discuss, unrelated to `Multi-var text`. If it might take a while, let's discuss it separately.
  - Currently, the text schema lacks proper overflow handling for boxes (always behaves like `overflow: visible`).
  - While developing the table schema, I added a [mechanism to page break](https://github.com/pdfme/pdfme/blob/40cc703f2c2ac144dfb395f9be3998f40b3ed43f/packages/common/src/helper.ts#L243C14-L243C32) (though there are still bugs).
  - How do you feel about adding a feature like `overflow: pageBreak` to the text schema?
  - If we can achieve `overflow: pageBreak`, it should improve the usability of `Multi-var text`.
    - This should make it even more user-friendly, as variable parts would break onto new pages when they're too long.
  - There are also table-related bugs to address before implementation. The timing for this is undecided; it's just an idea. I'd like to hear your thoughts.

peteward:
Hi @hand-dot,

Thanks for the feedback!

I have a deadline for the end of this week, but once I am passed this I can continue with this implementation. 

I agree that the additional outline in form mode is not ideal - this was just a quick visual proof of concept. I will think how we can make it better.

Page breaking is an interesting challenge. Let's consider this extension once the main implementation has been done. I can see more use-cases for this with the existing `text` schema, so hopefully we can find a solution that works for both.

Issue #474, opened 2024-04-26
Problem building project in NodeJS


mitcheljimenez:
### Describe the bug

I have a google cloud function project and I installed packages @pdfme/generator and @pdfme/common to generate PDFs inside our google cloud functions. When building the project, an error occurs because it cannot find namespace 'React' on packages like:

- antd
- @ant-design
- rc-picker
- form-render
- And others related to pdfme.

The weird thing is that I only installed those two packages and did not installed other packages like for designing pdf's, pdf viewer or pdf form.

![Captura de pantalla 2024-04-26 a la(s) 09 00 50](https://github.com/pdfme/pdfme/assets/28033507/cde68b13-0401-4d59-80e6-1ce0cc09d41e)

### To Reproduce

I created a template, installed @pdfme/common and @pdfme/generator and tried to generate pdf with some inputs.

![Captura de pantalla 2024-04-26 a la(s) 09 41 22](https://github.com/pdfme/pdfme/assets/28033507/c629ab2f-91fe-4458-a074-ae49ceff6f2c)

### Expected behavior

I would expect the packages to not need React on backend.

### Your Environment

```markdown
- pdfme package: @pdfme/generator & @pdfme/common
- pdfme version:
    "@pdfme/common": "^4.0.0",
    "@pdfme/generator": "^4.0.0",
- Operating system: MacOs
- Node.js version or Browser name & version: 20
```

### Your Error Log

```shell
node_modules/rc-virtual-list/lib/Filler.d.ts:1:24 - error TS7016: Could not find a declaration file for module 'react'. '/Users/mitcheljimenez/innovatech/insurance-carrier/functions/node_modules/react/index.js' implicitly has an 'any' type.
  Try `npm i --save-dev @types/react` if it exists or add a new declaration (.d.ts) file containing `declare module 'react';`

1 import * as React from 'react';
                         ~~~~~~~

node_modules/rc-virtual-list/lib/hooks/useScrollTo.d.ts:1:24 - error TS7016: Could not find a declaration file for module 'react'. '/Users/mitcheljimenez/innovatech/insurance-carrier/functions/node_modules/react/index.js' implicitly has an 'any' type.
  Try `npm i --save-dev @types/react` if it exists or add a new declaration (.d.ts) file containing `declare module 'react';`

1 import * as React from 'react';
                         ~~~~~~~

node_modules/rc-virtual-list/lib/interface.d.ts:1:23 - error TS2688: Cannot find type definition file for 'react'.

[...]

PR #477, opened 2024-04-30
fix(common): Move typing dependencies to peer

sebimarkgraf:
This could be a fix for #474 and the other related issues.

This ignores the type import of antd and casts it to any if antd is not installed.
Additionally, the antd dependency is downgraded to peer in common.

If ui is installed the direct dependency would lead to the correct typing being filled in.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -13804,10 +13804,12 @@
       "license": "MIT",
       "dependencies": {
         "@pdfme/pdf-lib": "^1.18.3",
-        "antd": "^5.11.2",
         "buffer": "^6.0.3",
-        "form-render": "^2.2.20",
         "zod": "^3.20.2"
+      },
+      "peerDependencies": {
+        "antd": "^5.11.2",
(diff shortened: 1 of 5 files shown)


sebimarkgraf:
Something that could be interesting, is doing the same thing with form-render, but I am unsure if form-render is a real dependency

sebimarkgraf:
The tests are passing with form-render as peer as well.

Should I do additional checks?

hand-dot:
Hey @sebimarkgraf !

Sorry for the delayed response. 
I plan to merge it if everything works fine on my end after verification.

svivenns:
Hi @hand-dot, any update on this MR, thank you.

hand-dot (approved):
LGTM!

hand-dot:
Hey @sebimarkgraf 

Sorry for the delayed review.
I will merge it now and release it in a patch version!

Thank you for your contribution!

hand-dot:
Released as V4.0.2!!

PR #494, opened 2024-07-12
Add strikethrough and underline formatter on text schema

necessarylion:
- Work on different font sizes
- Work on multiple lines
- Work on text with character spacing
- Work on text with align/vertical align

Designer

Screenshot 2024-07-12 at 11 25 38 AM <br>

Generated PDF

Screenshot 2024-07-12 at 11 25 54 AM


peteward:
Hi @necessarylion,

This is very cool, thanks for the contribution.

I think it's nearly all working except for character spacing is missing from the line length calculations:

Screenshot 2024-07-13 at 11 45 19
Screenshot 2024-07-13 at 11 45 42

But everything else that I've checked so far works.

I think for the UI panel it would be good to add the icons into an existing row, rather than adding a new one, perhaps on the left hand side of the Alignment options with a heading `Formatting`? We can make the other options thinner, or perhaps make text alignment buttons too then they would use less horizontal space. one for @hand-dot to comment on.

Also, I think it would be good to add an integration test for this formatting. I would be happy to help add this though after it is merged.

necessarylion:
Hi @peteward, thank you for the review. I have fixed character spacing issue [here](https://github.com/pdfme/pdfme/pull/494/commits/da9e5c65ac78f0ada2aec5e26f25358e843ac79c). 

Regarding the UI panel, alignment components were utilized across various schemas, whereas the new strikethrough and underline features are specific to the text schema. This is why I introduced them as new components. My initial idea was to place them next to the other alignment options, but I couldn't do so because they are shared components and also created as the custom widget for [form-render](https://www.npmjs.com/package/form-render).

If you have a better suggestion, please let me know; I'm happy to consider it.

peteward:
Thanks for the fast update @necessarylion .

I meant next to the _text_ alignment options (Horizontal: left, center, right & Vertical: bottom, middle, top), not the widget alignment options.

I might play around with this later today and see how it looks 👍

packages/ui/src/components/Designer/RightSidebar/DetailView/index.tsx
```diff
@@ -35,6 +36,7 @@ const DetailView = (
   useEffect(() => {
     const newWidgets: typeof widgets = {
       AlignWidget: (p) => <AlignWidget {...p} {...props} options={options} />,
+      TextFormatWidget: (p) => <TextFormatWidget {...p} {...props} options={options} />,
```
peteward, line 39: I can see what you have done here and why, but I think this needs to be more generic, and not tied to an implementation just for the text schema.

I think we could have a `BooleanButtonsWidget` which could be generic enough to pass in the names of attributes and SVG details to be rendered.

peteward: Using a quick test I could make this:

Screenshot 2024-07-16 at 06 32 12

I think we could do a similar buttons widget for text alignment too. I will try this later today.

necessarylion: Screenshot 2024-07-16 at 4 18 34 PM

@peteward How about like this?

And we can configure buttons like below

```ts
export enum Formatter {
  STRIKETHROUGH = 'strikethrough',
  UNDERLINE = 'underline',
  ALIGNMENT = 'alignment',
  VERTICAL_ALIGNMENT = 'verticalAlignment',
}


{
    title: i18n('schemas.text.format'),
    widget: 'ButtonGroup',
    buttons: [
      { key: Formatter.STRIKETHROUGH, svg: TextStrikethroughIcon, type: 'boolean' },
      { key: Formatter.UNDERLINE, svg: TextUnderlineIcon, type: 'boolean' },
      { key: Formatter.ALIGNMENT, svg: TextAlignLeftIcon, type: 'select', value: DEFAULT_ALIGNMENT },
      { key: Formatter.ALIGNMENT, svg: TextAlignCenterIcon, type: 'select', value: ALIGN_CENTER},
      { key: Formatter.ALIGNMENT, svg: TextAlignRightIcon, type: 'select', value: ALIGN_RIGHT},
      { key: Formatter.VERTICAL_ALIGNMENT, svg: TextVerticalAlignTopIcon, type: 'select', value: DEFAULT_VERTICAL_ALIGNMENT },
      { key: Formatter.VERTICAL_ALIGNMENT, svg: TextVerticalAlignMiddleIcon, type: 'select', value: VERTICAL_ALIGN_MIDDLE},
      { key: Formatter.VERTICAL_ALIGNMENT, svg: TextVerticalAlignBottomIcon, type: 'select', value: VERTICAL_ALIGN_BOTTOM},
    ]
    span: 16,
 };
```

necessarylion: I will push the code, so that you can check. And please feel free to give the feedback.

peteward: this is excellent, I really like it!

I had just played around with something similar but splitting the buttons into groups (but it was hard to fit one one line)

Screenshot 2024-07-16 at 10 51 09


Your code is better though and would give the flexibility to split them if desired 👍 

I think one final improvement would be to use SVGs that are more familiar to users working with text, like:

Screenshot 2024-07-16 at 10 54 38
Screenshot 2024-07-16 at 10 55 31

We could always add these later though.

@hand-dot, what do you think of this PR?

necessarylion: > I think one final improvement would be to use SVGs that are more familiar to users working with text.

Yeah agreed, but I was just following the same icon from upper align icons.

peteward: I understand.

Personally, I think it would look clearer to have different icons from the widget alignment ones to distinguish the functionality.

necessarylion: Screenshot 2024-07-16 at 8 19 01 PM

Updated.

peteward: great work!

packages/common/src/schema.ts
```diff
@@ -56,6 +56,7 @@ export const Dict = z.object({
   'schemas.text.max': z.string(),
   'schemas.text.fit': z.string(),
   'schemas.text.dynamicFontSize': z.string(),
+  'schemas.text.format': z.string(),
```
peteward, line 59: do we still need this?

necessarylion: Yes. It is for the [schema title](https://github.com/pdfme/pdfme/pull/494/files#diff-84fadaad049676d98756eacc6206f5f695bca78fe837522dd9b68bca10cb73a6R71-R72).

peteward: ah sorry, yes this is translations not schema definition 👍

packages/ui/src/components/Designer/RightSidebar/DetailView/index.tsx
hand-dot: yes yes! Really nice.

hand-dot:
@necessarylion 
I will find time to review and test as soon as possible. Please wait a moment.
Thank you for the excellent work.

@peteward 
I'm also very pleased with the improvements from a UX perspective.

hand-dot (approved):

hand-dot:
@necessarylion 
This feature is also backward compatible, so it's scheduled for release in the next version.
Thank you for your excellent work!!

@peteward 
If you have any additional comments, please let me know. If not, let's merge this.

peteward (approved):
Great!

PR #501, opened 2024-07-18
Fix playground height, fix spacing of text prop panel items

peteward:
Playground was causing the designer to scroll because the height wasn't set correctly.

Improve the spacing around the toolbars:

before:
Screenshot 2024-07-18 at 11 35 25

after:
Screenshot 2024-07-18 at 11 33 32


hand-dot:
@peteward 
Thank you always.

PR #502, opened 2024-07-18
Multi-variable Text Schema (again)

peteward:
I've reimplemented this to support the form view now as well.

Multi-variable text supports all existing text capabilities and:
* 0 to n variables
* can be used in place of 'read only text' (just don't add variables)
* can use the same variable more than once in text.

**Designer mode**

https://github.com/user-attachments/assets/4a4e50ce-5b43-4330-8612-02d1d5387415

**Form mode**

https://github.com/user-attachments/assets/58cac3de-a2a9-4109-8631-f34aca455539

Also works with other formatting options such as dynamic font sizing and rotation:
Screenshot 2024-07-18 at 15 17 25

TODO: 
- [ ] unit tests
- [ ] Integration tests


packages/common/src/types.ts
```diff
@@ -142,11 +150,13 @@ export interface PropPanel<T extends Schema> {
  * @property {function} pdf Function for rendering PDFs.
  * @property {function} ui Function for rendering UI.
  * @property {PropPanel} propPanel Object for defining the property panel.
+ * @property {function} [shouldRerenderVars] Function for determining which variables should be dependencies of the useEffect deciding whether to rerender the schema in the UI
  */
 export type Plugin<T extends Schema & { [key: string]: any }> = {
   pdf: (arg: PDFRenderProps<T>) => Promise<void> | void;
   ui: (arg: UIRenderProps<T>) => Promise<void> | void;
   propPanel: PropPanel<T>;
+  shouldRerenderVars?: (arg: RerenderCheckProps) => any[];
```
peteward, line 159: I couldn't find a better way to do this. 

When we are typing into the box we are assessing the content to see if a new variable needs to be created, which populates a variable box in the Right Details section. This requires an onChange to the schema, but by default this will lose focus whilst you are typing. 

I tried the value itself as one of the change factors on the `UseEffect` but this broke removing an image on the image schema.

Allowing the schema to determine whether it should re-render seemed like the best option

packages/schemas/src/multiVariableText/propPanel.ts
```diff
@@ -0,0 +1,139 @@
...
+        }
+      },
+
+    };
+  },
+  widgets: { ...parentPropPanel.widgets, mapDynamicVariables },
+  defaultSchema: {
+    ...parentPropPanel.defaultSchema,
+    type: 'multiVariableText',
+    icon: '<svg xmlns="http://www.w3.org/2000/svg" width="24" height="24" viewBox="0 0 24 24" fill="none" stroke="currentColor" stroke-width="2" stroke-linecap="round" stroke-linejoin="round" class="lucide lucide-type"><polyline points="4 7 4 4 20 4 20 7"/><line x1="9" x2="15" y1="20" y2="20"/><line x1="12" x2="12" y1="4" y2="20"/></svg>',
```
peteward, line 100: I used the standard `T` text icon.
I chose this because if you are using this schema then you don't need Read-Only Text, since it can support this functionality. Therefore, these 2 wouldn't be used together and the `T` is clear.

Screenshot 2024-07-18 at 15 31 00

packages/schemas/src/text/uiRender.ts
```diff
@@ -51,28 +33,120 @@ export const uiRender = async (arg: UIRenderProps<TextSchema>) => {
...
+        range.selectNodeContents(textBlock);
+        range.collapse(false); // Collapse range to the end
+        selection?.removeAllRanges();
+        selection?.addRange(range);
+      }
+    });
+  }
+};
+
+export const buildStyledTextContainer = async (arg: UIRenderProps<TextSchema>, value: string) => {
```
peteward, line 120: I have refactored this file into 2 functions that can be shared by the multi-variable text schema - `buildStyledTextContainer` and `makeElementPlainTextContentEditable`.

I think this makes this file a bit simpler and easy to read too.

packages/common/src/types.ts
hand-dot: @peteward 
I see that this value is being used as a dependency array in useEffect.
I believe it was added here after trying various things, but I wonder if it's difficult to achieve the same thing without adding this property by modifying or changing the image schema?

It's not that I'm particularly averse to adding new properties to the plugin type. In fact, I've recently come to think that icon and multilingual labels should be here, so I welcome the modification.

However, regarding`shouldRerenderVars`, I feel it's a complicated specification intended for people who understand the internal structure, so I'd prefer not to add it if possible.

peteward: Yeah I know, it's not ideal.

I will try again

peteward: Hi @hand-dot ,

I couldn't find a way to make it work for all scenarios without some configuration of the plugin. In my case there are changes to both the `value` (the text) and the `schema` (the variables) which need to be ignored when editing. 

For text entry plugins, if you are typing then you do not want the field to be rerendered until you leave edit mode. However, this is not the case for other schemas such as image and signature. 

I managed to simplify it so that instead of a function it's just a boolean which you can set on the plugin. It might be an advanced option, but it might help other plugin authors in future:

https://github.com/pdfme/pdfme/pull/502/commits/33c8992282f5f2f5cebadd5d54ba9b554e519dbb

Let me know if you think this is better.

hand-dot:
Hey @peteward,
I'm ready to merge this PR whenever you are. The main addition here is the new schema, and I've confirmed we can release it without breaking backward compatibility since it doesn't significantly alter the common components.

I know you've been working on implementing this idea for a while now, and I've been excited to see it come to fruition. 
You've done an outstanding job!!

Just merge it when you're ready, and I'll take care of the release right away.
Thanks again for your hard work on this.

hand-dot (approved):

peteward:
Amazing, thank you!

I will write some documentation for this schema in the coming days 👍

packages/schemas/src/multiVariableText/propPanel.ts
snax4a: May I ask what is this `icon` property in schema definitions used for? Could not find it in the docs.

Issue #497, opened 2024-07-18
When a large number of plugins are set, the left sidebar of the designer overflows.


hand-dot:
### Describe the bug
![Untitled](https://github.com/user-attachments/assets/8e9257b8-968e-4ff4-aebb-b723c4b1aaa8)

### To Reproduce

Set a large number of plugins to the designer

### Expected behavior

Make it scrollable

### Your Environment

```markdown
- pdfme package(@pdfme/generator or @pdfme/ui): @pdfme/ui
- pdfme version: latest 
- Operating system: mac
- Node.js version or Browser name & version: latest chrome
```

### Your Error Log

```shell
none
```

hand-dot:
I'm sorry. I forgot to check who was assigned to the issue....

PR #506, opened 2024-07-20
Fix - When a large number of plugins are set, the left sidebar of the designer overflows.

hand-dot:
REF: https://github.com/pdfme/pdfme/issues/497

Before: 
image

After: 
CleanShot 2024-07-20 at 16 55 02


peteward:
ha! I had just done this 😄 

Check [my solution](https://github.com/pdfme/pdfme/pull/506/) it puts the scrollbar on the outside

hand-dot:
@peteward 
Hello, good morning.
Understood. I'll review your modifications and adopt the one that seems best. Please give me a moment.

peteward:
Yes, let's go with yours 👍

PR #508, opened 2024-07-21
Enforce validation: Do not apply schema changes from right sidebar if they are invalid

peteward:
I think this is a decent solution to #354. Originally I was thinking of trying to validate a schema fully on-demand, but I think this approach is a good start and perhaps enough.

In my company, we have problems where users enter and save invalid colors and the rendering fails from their schema. This change prevents invalid attribute values from getting into the schema in the first place, and the most likely cause of these errors is from human input via the designer. 

BEFORE:
You edit a value, change to another field, come back to the previous field and it doesn't show up as invalid.

https://github.com/user-attachments/assets/a4fd6558-8875-4738-a12d-ad35c2564b17

AFTER:

https://github.com/user-attachments/assets/432bb01c-c8aa-4e8b-a3b9-78b032f1b68d

NOTE: There are 2 caveats:
1. When you hover over another schema in the designer it appears to trigger a rerender of the right-hand sidebar and the value is therefore reset. I don't think this is a problem, perhaps a minor bug to look at in the designer.
2. Not every form field will map exactly to the schema keys, but I think in nearly every case it does.

### DIFF
--- a/packages/schemas/src/constants.ts
+++ b/packages/schemas/src/constants.ts
@@ -1,2 +1,2 @@
 export const DEFAULT_OPACITY = 1;
-export const HEX_COLOR_PATTERN = '^#(?:[A-Fa-f0-9]{3,4}|[A-Fa-f0-9]{6}|[A-Fa-f0-9]{8})$';
+export const HEX_COLOR_PATTERN = '^#(?:[A-Fa-f0-9]{6})$';
--- a/packages/ui/src/components/Designer/RightSidebar/DetailView/index.tsx
+++ b/packages/ui/src/components/Designer/RightSidebar/DetailView/index.tsx
@@ -1,6 +1,6 @@
 import FormRender, { useForm } from 'form-render';
 import React, { useContext, useEffect, useState } from 'react';
-import type { Dict, SchemaForUI, PropPanelWidgetProps, PropPanelSchema } from '@pdfme/common';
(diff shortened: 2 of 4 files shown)


packages/schemas/src/constants.ts
```diff
@@ -1,2 +1,2 @@
 export const DEFAULT_OPACITY = 1;
-export const HEX_COLOR_PATTERN = '^#(?:[A-Fa-f0-9]{3,4}|[A-Fa-f0-9]{6}|[A-Fa-f0-9]{8})$';
+export const HEX_COLOR_PATTERN = '^#(?:[A-Fa-f0-9]{6})$';
```
peteward, line 2: the renderer colour formatting only accepts full 6 char hex codes, nothing else works (currently)

packages/ui/src/components/Designer/RightSidebar/DetailView/index.tsx
```diff
@@ -75,22 +76,39 @@ const DetailView = (
...
 
   const handleWatch = (newSchema: any) => {
-    const changes = [];
+    let changes: ChangeSchemaItem[] = [];
     for (let key in newSchema) {
       if (['id', 'content'].includes(key)) continue;
+
+      // [position] Return the flattened position to its original form.
+      if (key === 'x') key = 'position.x';
+      if (key === 'y') key = 'position.y';
```
peteward, line 85: we need to flatten before we check if the value has changed, previously this was comparing against undefined so we always considered `x` and `y` had changed.

hand-dot (approved):
LGTM!

PR #510, opened 2024-07-21
Fix dynamic resizing additional newline

peteward:
`innerText` was returning the value with an additional `\n` on it. 

This was creating an extra line in read-only mode which was making the text shorter than the available space:

`textContent` does not have this additional `\n` on the end.

BEFORE:

https://github.com/user-attachments/assets/1a5cb676-a655-49f1-b214-9b5688efb374


hand-dot (approved):

hand-dot:
Released as 4.2.2!

PR #514, opened 2024-07-23
Fix schema change attribute retention

### LINKED ISSUE #513: When changing the field type, the x and y coordinates are changed to 0.
### Describe the bug

CleanShot 2024-07-23 at 10 04 50

This must have happened due to recent changes. It only occurs in the local playground and does not occur in the [production playground](https://playground.pdfme.com/).

### To Reproduce

Edit Field -> Change the field type
CleanShot 2024-07-23 at 10 11 00@2x

### Expected behavior

Do not change the position.

### Your Environment

```markdown
- pdfme package(@pdfme/generator or @pdfme/ui): @pdfme/ui
- pdfme version: latest
- Operating system: Mac
- Node.js version or Browser name & version: latest chrome
```

### Your Error Log

```shell
none
```

### Additional context

_No response_

peteward:
fixes #513

### DIFF
--- a/packages/ui/src/components/Designer/RightSidebar/DetailView/index.tsx
+++ b/packages/ui/src/components/Designer/RightSidebar/DetailView/index.tsx
@@ -75,21 +75,34 @@ const DetailView = (
     form.setValues(values);
   }, [form, activeSchema]);
 
-  const handleWatch = (newSchema: any) => {
+  const handleWatch = (formSchema: any) => {
+    const formAndSchemaValuesDiffer = (formValue: any, schemaValue: any): boolean => {
+      if (typeof formValue === 'object') {
+        return JSON.stringify(formValue) !== JSON.stringify(schemaValue);
+      }
+      return formValue !== schemaValue;
+    }
+
     let changes: ChangeSchemaItem[] = [];
-    for (let key in newSchema) {
+    for (let key in formSchema) {
       if (['id', 'content'].includes(key)) continue;
 
-      // [position] Return the flattened position to its original form.
-      if (key === 'x') key = 'position.x';
-      if (key === 'y') key = 'position.y';
+      let value = formSchema[key];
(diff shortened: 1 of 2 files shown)


packages/ui/src/components/Designer/RightSidebar/DetailView/index.tsx
```diff
@@ -75,21 +75,34 @@ const DetailView = (
     form.setValues(values);
   }, [form, activeSchema]);
 
-  const handleWatch = (newSchema: any) => {
+  const handleWatch = (formSchema: any) => {
+    const formAndSchemaValuesDiffer = (formValue: any, schemaValue: any): boolean => {
+      if (typeof formValue === 'object') {
+        return JSON.stringify(formValue) !== JSON.stringify(schemaValue);
```
peteward, line 81: this is needed otherwise comparisons of objects like `dynamicFontSize` would always be false and therefore always considered as changed.

packages/ui/src/helper.ts
```diff
@@ -454,10 +454,15 @@ const handleTypeChange = (
       delete schema[key as keyof typeof schema];
     }
   });
+  // Apply attributes from new defaultSchema
   const propPanel = Object.values(pluginsRegistry).find(
     (plugin) => plugin?.propPanel.defaultSchema.type === value
   )?.propPanel;
-  Object.assign(schema, propPanel?.defaultSchema || {});
+  Object.keys(propPanel?.defaultSchema || {}).forEach((key) => {
```
peteward, line 461: this was the real bug - we were saving the attributes we wanted above but then overwriting them with the defaults from the new schema (which includes `position { x: 0, y: 0 }`).

I think this worked before because of the order of events. 

It triggered change on `type` first, then (accidentally) triggered change on `x` and `y` so the values were reset to 0 but then quickly put back to what they were before.

PR #521, opened 2024-07-25
Add a required attribute to fields which is validated by the designer

peteward:
It would need documentation and tests, but this is working.

I avoided adding `required: false` to the `defaultSchema` of every schema type and instead handled it in the Designer when new fields are added

I don't love how this looks (the default checkbox layout with a horizontal label)...

Screenshot 2024-07-25 at 15 03 42


peteward:
Let me know what you think of this approach @hand-dot .

If you don't want this to be default functionality, then the `validateRequiredFields` could be passed in as a preprocessing function, but I would still need the `required` functionality within the UI. Although, this could also be hidden with `options` if necessary.

hand-dot:
@peteward 

I think it's a great idea. 
I like the concept, but I'll provide feedback on thoughts I had while running it in the playground.

- I always want the Form, Viewer, and the actual created PDF to match, but I feel uncomfortable that there's no feedback when required fields in the Form and Viewer are not filled in.
- It might be better to make it more easily understandable to users, such as adding a "*" mark to the top left of the schema or change border.(same as Form)
  -  CleanShot 2024-07-27 at 10 01 27@2x
- This may be slightly off-topic, but it might be good to implement ReadOnly in the same way as this Required checkbox.
  - CleanShot 2024-07-27 at 09 59 41@2x
- Would it be easier for users to use the designer if the Field List in the designer also captured and displayed schema information such as required, readonly, and type?
  - CleanShot 2024-07-27 at 10 08 32@2x

peteward:
These are great points and ideas!

Let me review and try to apply them then I will update the PR.

hand-dot:
Thank you.
Perhaps for v5, we might consider making plugin icons mandatory rather than optional to maintain design consistency.

peteward:
Updated:

ListView now shows icons and required `*` in Designer:

Screenshot 2024-07-29 at 14 29 10

Form view includes `*` before fields if required

Screenshot 2024-07-29 at 14 28 54

Added a simple alert to the playground for catching generation errors:

Screenshot 2024-07-29 at 14 30 52

I haven't looked at read-only though, I think that could be done in a separate PR because there's more to think about with that change.

packages/schemas/src/text/propPanel.ts
```diff
@@ -39,9 +39,12 @@ const UseDynamicFontSize = (props: PropPanelWidgetProps) => {
     changeSchemas([{ key: 'dynamicFontSize', value: val, schemaId: activeSchema.id }]);
   };
   const label = document.createElement('label');
-  label.innerText = i18n('schemas.text.dynamicFontSize') || '';
+  const span = document.createElement('span');
+  span.innerText = i18n('schemas.text.dynamicFontSize') || '';
+  span.style.cssText = 'margin-left: 0.5rem';
```
peteward, line 44: I change this so that the checkbox is on the left side of the label, to match the antd layout that is used for `Required`.

I couldn't find an easy way to use a proper checkbox widget from antd though and make the functionality work though.

hand-dot: Nice.

packages/generator/src/helper.ts
```diff
@@ -58,6 +58,16 @@ export const getEmbedPdfPages = async (arg: { template: Template; pdfDoc: PDFDoc
   return { basePages, embedPdfBoxes };
 };
 
+export const validateRequiredFields = (template: Template, inputs: Record<string, any>[]) => {
+  template.schemas.forEach((schemaObj) =>
+    Object.entries(schemaObj).forEach(([fieldName, schema]) => {
+      if (schema.required && !schema.readOnly && !inputs.some((input) => input[fieldName])) {
+        throw new Error(`input for '${fieldName}' is required to generate this PDF`);
```
hand-dot, line 65: `[@pdfme/generator] input for '${fieldName}` is better

packages/generator/src/generate.ts
```diff
@@ -2,17 +2,19 @@ import * as pdfLib from '@pdfme/pdf-lib';
...
+import { insertPage, preprocessing, postProcessing, getEmbedPdfPages, validateRequiredFields } from './helper.js';
 
 const generate = async (props: GenerateProps) => {
   checkGenerateProps(props);
   const { inputs, template, options = {}, plugins: userPlugins = {} } = props;
   const basePdf = template.basePdf;
 
   if (inputs.length === 0) {
-    throw new Error('inputs should not be empty');
+    throw new Error('inputs should not be empty, pass at least an empty object in the array');
```
hand-dot, line 13: Could you add the following prefixes that users can distinguish between user-caused errors and system-caused errors?

 `[@pdfme/generator]` .

packages/ui/src/components/Designer/PluginIcon.tsx
hand-dot, line 1: Cool!

packages/ui/src/components/Designer/RightSidebar/DetailView/index.tsx
```diff
@@ -119,6 +119,7 @@ Check this document: https://pdfme.com/docs/custom-schemas`);
         span: 12,
       },
       key: { title: i18n('fieldName'), type: 'string', required: true, span: 12 },
+      required: { title: i18n('required'), type: 'boolean', span: 6, hidden: defaultSchema?.readOnly },
```
hand-dot, line 122: Ultimately, I want to eliminate the read-only schema and make it toggleable with a checkbox, similar to the Required option. 

I think it's fine to address this in a separate PR.

packages/ui/src/components/Designer/RightSidebar/ListView/Item.tsx
hand-dot, line 1: ❤️

packages/ui/src/i18n.ts
hand-dot, line 1: Thank you ChatGPT. 

Looks good. At least Japanese, perfect!

packages/ui/src/helper.ts
```diff
@@ -463,6 +463,9 @@ const handleTypeChange = (
       (schema as any)[key] = propPanel?.defaultSchema[key];
     }
   });
+  if (schema.readOnly) {
```
hand-dot, line 466: [option]  
If we can use  `options.requiredByDefault`,  `schema.required` should be determine by this option.

hand-dot:
@peteward 
- No need for asterisks (*) on read-only schemas in forms

https://github.com/user-attachments/assets/f8864411-cbd2-4a9b-9a75-d258833f150d

- I think it would be better to have asterisks (*) in the upper right corner of the designer's schema as well

packages/generator/src/generate.ts
peteward: Yes, I will add this.

I wonder if in future we might add a `PdfmeError` class extending `Error` where we can separate the `package` (`generator`, `ui`, `common`, etc) as an attribute. 

This way you could separate the actual error message from the service so that if you did want to expose a message to end users rather than developers you could.

Something to consider for future maybe.

packages/ui/src/components/Designer/RightSidebar/ListView/SelectableSortableContainer.tsx
```diff
@@ -136,27 +146,17 @@ const SelectableSortableContainer = (
...
-                    .map((item) => (
-                      <Item
-                        key={item.id}
-                        value={item.key}
-                        style={{ background: token.colorPrimary }}
-                        dragOverlay
-                      />
-                    ))}
-                </ul>
-              </>
```
peteward, line 159: Oh, I forgot to mention, I didn't know what this was achieving so I removed it?

This seemed to create draggable list items when multiple are being moved together? (I think) but I couldn't see how this was possible. If you select multiple schemas then it always goes to the `DetailView`. I couldn't see a way to get multiple draggable schemas in the `ListView`. 

If am I wrong then I will revert this. If it's correct then there might be more code that can be removed from this file.

packages/ui/src/helper.ts
peteward: this is handled on [line 174 of `Designer/index.tsx` ](https://github.com/pdfme/pdfme/blob/pw-required-fields/packages/ui/src/components/Designer/index.tsx#L175)- when a new field is added it is set here. A readOnly schema will always not be required.

peteward:
I have updated:

* Show `*` on designer
* Avoid setting required for readOnly schemas (until we make this toggleable)
* Error name prefixes
* Removed the `requiredByDefault` setting in the playground example

I have spotted a bug with the required checkbox not clearing when you switch schemas. I will fix this later this morning then update with fixed tests.

packages/generator/src/generate.ts
hand-dot: 100% ggree!

packages/ui/src/components/Designer/RightSidebar/ListView/SelectableSortableContainer.tsx
hand-dot: I had made it so that pressing shift while selecting would allow multiple items to be moved at once, but it seems to have stopped working at some point.

It might be better not to delete it.

peteward: OK, I will revert

packages/ui/src/helper.ts
hand-dot: Ah, Okay👍

packages/ui/src/components/Designer/RightSidebar/ListView/SelectableSortableContainer.tsx
peteward: I have reverted this.

peteward:
OK, I think this is ready now 🤞

hand-dot:
Okay, I will review and test this from now.

hand-dot (approved):
Very nice! It's ready to merge. I'd like @peteward to merge it whenever you wants to.

I want to confirm about the release. If the PR to add the readonly checkbox can be implemented quickly, I'd prefer to release after that's done.

peteward:
OK. 

If you want me to look at reworking `readOnly` then it might have to be later in the week, but I will try to find time.

PR #523, opened 2024-07-26
Fewer rerenders of the DetailView in RightSidebar

peteward:
When debugging a separate issue I noticed that the `RightSidebar` `DetailView` was being rendered whenever I hovered over other schemas. This should not be necessary because the selected schema has not changed, just hovering over something else shouldn't trigger it to re-render.

I investigated this and found that the props being passed into the DetailView were being considered as changed, and therefore the whole component is being rendered again. I debugged these props though and they weren't changing...
However, it seems that when a prop is a function or object, if the reference to these objects has changed then it will also cause any child components using them to be re-rendered.

This component (and I think there are others that are similar) is designed to be long-running, with 2 `useEffect` calls looking for changes. However, what is actually happening is the whole component gets re-rendered anyway, so `useEffect` calls are unnecessary.

This improvement uses `React.memo()` with a check for whether the component should really be re-rendered, and that is only if the `activeSchema` has changed from the previously passed in `activeSchema`. This avoids any re-rendering on hovering over other schemas, and also avoids a double re-render when changes are made in the sidebar.  It does not prevent a rerender after interacting with the DetailView, but I can't think of a way of avoiding this. If you make a change in the DetailView, it changes the schema, which then triggers the DetailView to rerender. You would need to check whether the values in the form match those attributes in the schema and if they are the same then don't rerender, which is non-trivial!

I think this is a common pattern that we might see in the UI where components are designed with long-living intentions and `useEffect()` but in reality they get fully re-rendered anyway. There might be a better architectural solution to the wider issue, or we should consider using this `React.memo` pattern in other areas.

BEFORE:

https://github.com/user-attachments/assets/82d8deeb-2451-47c9-9b52-85953a1e1d46

AFTER:

https://github.com/user-attachments/assets/f064f5e1-5460-4747-8305-b1925d8f7a3a

### DIFF
--- a/packages/ui/src/components/Designer/RightSidebar/DetailView/index.tsx
+++ b/packages/ui/src/components/Designer/RightSidebar/DetailView/index.tsx
@@ -13,14 +13,13 @@ import { InternalNamePath, ValidateErrorEntity } from "rc-field-form/es/interfac
 
 const { Text } = Typography;
 
-const DetailView = (
-  props: Pick<
-    SidebarProps,
+type DetailViewProps = Pick<SidebarProps,
     'size' | 'schemas' | 'pageSize' | 'changeSchemas' | 'activeElements' | 'deselectSchema'
-  > & {
-    activeSchema: SchemaForUI;
-  }
-) => {
+> & {
+  activeSchema: SchemaForUI;
+};
+
+const DetailView = (props: DetailViewProps) => {
   const { token } = theme.useToken();
 
   const { size, changeSchemas, deselectSchema, activeSchema, activeElements } = props;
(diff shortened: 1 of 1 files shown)


hand-dot (approved):

hand-dot:
@peteward 

Thank you. I understand that re-rendering issues are indeed troublesome, and the current pdfme/ui is not very well optimized.

Personally, I'm looking forward to React Compiler, which will be introduced in React 19.
- https://react.dev/blog/2024/02/15/react-labs-what-we-have-been-working-on-february-2024#react-compiler
- https://react.dev/learn/react-compiler

I believe fixes are necessary when problems are clear, but personally, I don't want to overwork the code for perfect optimization at this stage.

I will include this fix in the next release. Thank you as always.

PR #525, opened 2024-07-28
Fixed and added add row and delete row functionlity

rauni00:
Now able to add row and delete that row


rauni00:
Now able to add row and delete that row

rauni00:
In the table

hand-dot:
@rauni00 
Thank you for sending the pull request.
So, this PR enables adding and deleting rows in the table on the designer, is that right? At this point, I don't think this feature is necessary.

When I initially designed it, I didn't think it was necessary to be able to add or delete rows there, as the designer is for configuring the table's design settings.

The actual number of rows is determined by the data passed to the generate function.
Given the above, could you tell me if there are any more reasons why this might be necessary?

rauni00:
@hand-dot 
this PR enables adding and deleting rows in the table on the designer, is that right -- **Yes**
Yes it is necessary because i'm creating a items list in the table but not able to add more list in the table able to add row 
But primary i need more item list in the table how can do this is there another way to do ?? 

The actual number of rows is determined by the data passed to the generate function- Yes this working but how to do in editor
Please tell

hand-dot:
@rauni00 
It is not possible to determine the rows of the table on the designer.
This is because the designer is for creating template definitions, while the actual PDF creation is done by the generator package.

Therefore, while the designer determines the settings for each schema, it doesn't need to know how the actual PDF will be created or the parts that depend on the data.

For example, an approach that allows setting min-max limits for the number of rows in the table might be the correct way to address what you're trying to do.

hand-dot:
With the following issue, we will add the readOnly attribute to the schema, making it possible to make all schemas non-editable. Therefore, the possibility of including this feature to display tables with fixed data in the implementation has increased.
https://github.com/pdfme/pdfme/issues/541

At this point, it's still undecided, so we'll keep this PR on hold.

hand-dot:
ref: https://github.com/pdfme/pdfme/pull/595

PR #533, opened 2024-08-01
[Text schema]: Add missing color type in get font prop

necessarylion:
Since color type was missing in `getFontProp` function, setting color type option was not working and always rendered as RGB in v4.


hand-dot (approved):
Oh... Thank you so much!!

Issue #529, opened 2024-07-31
Changing a field name so that it duplicates another will remove one of the fields


peteward:
### Describe the bug

If you change the name of a field to that of an existing one then one of the fields will be removed as soon as you select off of the schema.

### To Reproduce

* Add a schema `field1`
* Add another schema and rename it to `field1` 
* When you lose focus on that schema there will only be 1 field remaining

### Expected behavior

Now that we have code that prevents field updates if they are invalid, we should be able to apply a custom validation rule to the field name form element which will mark it as an error if it equals any other field names.  This will prevent it being saved, so no fields will be accidentally 'lost'.

### Your Environment

```markdown
- pdfme package(@pdfme/generator or @pdfme/ui): UI
- pdfme version: latest main
```

### Your Error Log

```shell
-
```

hand-dot:
@peteward 
I can't reproduce...
Could you update your Chrome?

Chrome: Version 127.0.6533.89 (Official Build) (arm64)

![CleanShot 2024-07-31 at 18 19 01](https://github.com/user-attachments/assets/ba1b0643-c534-493d-a7d1-631d3fcf4568)

peteward:
Hmm interesting...

Version 127.0.6533.89 (Official Build) (arm64)


https://github.com/user-attachments/assets/ddc7a43b-c747-4e64-94b9-75fbd66c0791

hand-dot:
Hmm.
Does it also occur in an incognito browser on the [playground](https://playground.pdfme.com)?
There's a possibility that Chrome extensions might be involved.

If it does occur in an incognito browser, could it be due to this fix?
https://github.com/pdfme/pdfme/pull/506/files

This might be a bit of a leap, but I'd like to make it possible to automatically update the PDFME_VERSION below, in order to check which version the current playground is running on.
https://github.com/pdfme/pdfme/blob/main/packages/common/src/constants.ts#L2

peteward:
It seems to be due to the Lastpass chrome extension.

I think the improvement to validation is still worth exploring, but I will also try `data-lpignore="true"` to see if this prevents it.

peteward:
implemented

Issue #528, opened 2024-07-31
Adding new field names should be more intelligent


peteward:
### Describe the bug

This is an edge case, but if you add your first field and rename it to `field2` then any other field you add removes the previous field :D 

### To Reproduce

https://github.com/user-attachments/assets/f651e8b4-84bf-4a4b-9bdd-8d2f94cc0fec

### Expected behavior

Field naming when creating a new field should consider existing fields and find a name that is not used.

### Your Environment

```markdown
- pdfme package(@pdfme/generator or @pdfme/ui): UI
- pdfme version: latest main
```

### Your Error Log

```shell
-
```

PR #540, opened 2024-08-03
Validating unique schema keys

peteward:
implements #528 #529

1. When a new schema is created ensure the new key is unique

https://github.com/user-attachments/assets/7487e111-504e-4d8f-b196-a725cc2680fd

2. Validate a schema key when editing it and do not save on collisions, retain the last valid value

https://github.com/user-attachments/assets/aca63af1-42e3-4877-ae67-a90e59a86e68


packages/ui/src/components/Designer/RightSidebar/DetailView/index.tsx
```diff
@@ -66,14 +66,29 @@ const DetailView = (props: DetailViewProps) => {
     values.y = values.position.y;
     delete values.position;
 
-    if (values.key !== (form.getValues() || {}).key) {
```
peteward, line 69: move this to the `useEffect` below due to an edge case. If you change the key to that of another schema then select the new schema of the same key it wouldn't reset.

packages/ui/src/components/Designer/RightSidebar/DetailView/index.tsx
```diff
@@ -66,14 +66,29 @@ const DetailView = (props: DetailViewProps) => {
...
+      }
+      return true;
+    };
+  }, [schemas, activeSchema]);
+
+  const uniqueSchemaKey = useRef((value: string): boolean => true);
+
+  const validateUniqueSchemaKey = (_: any, value: string): boolean => {
+    return uniqueSchemaKey.current(value);
+  }
```
peteward, line 90: This was the only way I could get the `schemas` variable to update.
Otherwise it used a stale value of `schemas` from first page load. I tried wrapping the function in `useCallback` but this didn't work...

hand-dot (approved):
LGTM!

PR #544, opened 2024-08-06
multiVariableText default schema sizes and text

peteward:
After feedback from my team internally they felt that this change in default text would help users better understand how to use multi variable text (and made the field bigger to ensure the content fits with most fonts)


hand-dot:
Released as 4.2.5!

Issue #541, opened 2024-08-03
Integrate the ReadOnly schema


hand-dot:
### Is your feature request related to a problem? Please describe.

Until now, the ReadOnly schema and normals were separated into different schemas, but we will integrate them. We will make it possible to set ReadOnly from a checkbox, similar to how the 'required' checkbox is set.

- ref: https://github.com/pdfme/pdfme/pull/521#pullrequestreview-2207699785

### Describe the solution you'd like

- Add ReadOnly checkbox
  - ![image](https://github.com/user-attachments/assets/b7e46ca6-fe84-419f-befc-2d08fcbdf842)
- Integrate the ReadOnly schema

### Describe alternatives you've considered

None

### Additional context

This modification should be a major version update if we're going to delete the existing ReadOnly schema. It might be better to keep the ReadOnly schema for a while, but with this change, ReadOnly schema will eventually become unnecessary, so it's planned to be deprecated in the future.

peteward:
I'm wondering if there's a better (English) word to use instead of "read only". I don't know if this would be clear to an editor.
Maybe **static** vs **dynamic** or **fixed** vs **editable**. But I guess this is just a label, and at the code level it's fine to be `readOnly`

Mystificious:
> I'm wondering if there's a better (English) word to use instead of "read only". I don't know if this would be clear to an editor. Maybe **static** vs **dynamic** or **fixed** vs **editable**. But I guess this is just a translation label, and at the code level it's fine to be `readOnly`

Maybe "Not editable" ?

hand-dot:
> Maybe static vs dynamic or fixed vs editable. 

> Maybe "Not editable" ?

I can't decide on my own. @peteward I want to trust your judgment.

hand-dot:
If @peteward hasn't started working on it yet, would it be okay if I implement it myself?

peteward:
Of course. 

Yes, I haven't started. 👍

dkastl:
I actually did this in a project, where I extended some schemas with additonal fields. These additonal fields are not relevant for PDFme, but I felt adding a "switch" widget for setting the read-only flag was relatively easy:

https://github.com/gtt-project/redmine_print_templates/blob/next/src/plugins/extended/schemaUtils.ts#L47-L52

It would be nice if integrating the ReadOnly setting into PDFme. I actually asked the question once in Discord.

dkastl:
Thanks a lot for adding this functionality!

Issue #542, opened 2024-08-03
Utils of @pdf/schemas not exported properly


Mystificious:
### Is your feature request related to a problem? Please describe.

Greetings Kyohei,

I am trying to implement some custom schemas (inputs and checkboxes that keep editable in the generated PDF -> Creating PDF-Forms with pdfme) and want to use the functions of the utils.ts but those are not exported. Some of the provided utils are pretty handy for use with custom schemas. e.g. the Color-transformations.

Same goes for the helpers.ts of the most complex schema of "text"... There are a lot of needy translations you can use for other schemas as well.

Is there a reason why you aren't exporting all functions from utils? That would lead to a more standardised way of creating custom schemas for the whole community.

All the Best from Berlin!

### Describe the solution you'd like

packages/schemas/src/index.ts:

```
import multiVariableText from './multiVariableText/index.js'
import text, { readOnlyText } from './text/index.js';
import image, { readOnlyImage } from './graphics/image.js';
import svg, { readOnlySvg } from './graphics/svg.js';
import barcodes from './barcodes/index.js';
import line from './shapes/line.js';
import table from './tables/index.js';
import { modifyTemplateForTable, getDynamicHeightForTable } from './tables/dynamicTemplate.js';
import { rectangle, ellipse } from './shapes/rectAndEllipse.js';
import { convertForPdfLayoutProps, rotatePoint, hex2PrintingColor, hex2CmykColor, hex2RgbColor, isEditable, readFile, createErrorElm } from './utils.js';
import {getBrowserVerticalFontAdjustments, getFontDescentInPt, heightOfFontAtSize, widthOfTextAtSize, getFontKitFont, getSplittedLines, calculateDynamicFontSize} from '../src/text/helper';


const tableBeta = table;

const builtInPlugins = { Text: text };

export {
  multiVariableText,
  text,
  readOnlyText,
  image,
  readOnlyImage,
  svg,
  readOnlySvg,
  barcodes,
  line,
  tableBeta,
  modifyTemplateForTable,
  getDynamicHeightForTable,
  rectangle,
  ellipse,
  builtInPlugins,
  convertForPdfLayoutProps,
  rotatePoint,
  hex2PrintingColor,
  hex2CmykColor,
  hex2RgbColor,
  isEditable,
  readFile,
  createErrorElm,
  getBrowserVerticalFontAdjustments,
  getFontDescentInPt,
[...]

PR #547, opened 2024-08-07
Integrate the ReadOnly schema

hand-dot:
REF: https://github.com/pdfme/pdfme/issues/541

CleanShot 2024-08-07 at 11 25 37@2x

CleanShot 2024-08-07 at 11 26 21@2x


packages/ui/src/components/Designer/RightSidebar/DetailView/index.tsx
```diff
@@ -174,9 +174,10 @@ Check this document: https://pdfme.com/docs/custom-schemas`);
           validator: validateUniqueSchemaKey,
           message: i18n('validation.uniqueName'),
         }],
-        props: { autocomplete: "off" }
+        props: { autoComplete: "off" }
       },
-      required: { title: i18n('required'), type: 'boolean', span: 8, hidden: defaultSchema?.readOnly },
+      required: { title: i18n('required'), type: 'boolean', span: 8, hidden: defaultSchema?.readOnly ,disabled: '{{formData.readOnly}}',},
```
hand-dot, line 179: `{{formData.readOnly}}` : It is not possible to set both ReadOnly and Required to true simultaneously.

packages/ui/src/components/Designer/RightSidebar/ListView/SelectableSortableContainer.tsx
```diff
@@ -152,31 +151,39 @@ const SelectableSortableContainer = (
...
-                    required={schemas.find((schema) => schema.id === activeId)!.required}
-                    style={{ background: token.colorPrimary }}
-                    dragOverlay
-                  />
-                </ul>
-                <ul style={{ margin: 0, padding: 0, listStyle: 'none' }}>
-                  {selectedSchemas
-                    .filter((item) => item.id !== activeId)
-                    .map((item) => (
+              (() => {
```
hand-dot, line 154: Just refactoring.

packages/schemas/src/tables/uiRender.ts
hand-dot, line 1: If read-only is supported in tables, the feature added in the PR below that allows adding fixed data in the designer might be useful.
https://github.com/pdfme/pdfme/pull/525

hand-dot:
@peteward Could you check this PR?

packages/ui/src/components/Designer/RightSidebar/DetailView/index.tsx
peteward: I think we might need a new attribute on the plugin, something like `supportsReadOnly`?  🤔 

For `multiVariableText` it handles readOnly implicitly by whether you include any `{variables}` in your text. If no variables then it's effectively readOnly, otherwise it is not.

packages/ui/src/i18n.ts
```diff
@@ -11,6 +11,7 @@ const dictEn: { [key in keyof Dict]: string } = {
   opacity: 'Opacity',
   rotate: 'Rotate',
   required: 'Required',
+  readOnly: 'Not editable',
```
peteward, line 14: I'm still trying to think of the best description. I will keep thinking...

peteward:
will this PR also remove the dedicated read only schema definitions for image, text ,SVG, etc?

packages/ui/src/i18n.ts
peteward: I'm starting to think that just `Editable` is the best UX, ticked by default for those fields that support it (which would mean either inverting the saved setting value or changing the schema structure). 

I will get some feedback from colleagues later today.

hand-dot:
@peteward Thank you for reviewing this PR.

>  will this PR also remove the dedicated read only schema definitions for image, text ,SVG, etc?

No, this PR does not remove the read-only schemas.
However, they will become unnecessary in the future, so I'm considering removing those schemas or making them importable from @pdfme/schemas/deprecated when we do a major update like V5.
ref: https://github.com/pdfme/pdfme/issues/542

packages/ui/src/components/Designer/RightSidebar/DetailView/index.tsx
hand-dot: Hmm, I see. I understand what you're saying.

However, multiVariableText is an exception, and in most cases, it's generally easy to support read-only functionality. (I had overlooked multiVariableText's such  behavior.)

If we can avoid adding a new property by modifying multiVariableText, I think we should do that. What do you think?

packages/ui/src/i18n.ts
hand-dot: Is "Read Only" really that problematic? I understand that my opinion doesn't carry much weight here since I can only judge from a Japanese language perspective, but at least in Japanese, it doesn't feel awkward at all.

I apologize for making a statement that brings the discussion back to square one.

packages/ui/src/components/Designer/RightSidebar/DetailView/index.tsx
peteward: I can't think of a solution now, but I will keep thinking

packages/ui/src/i18n.ts
peteward: No problem! :) 

After speaking with a couple of people, we think that this is the best (English) UX:

https://github.com/user-attachments/assets/99542e0c-cad0-491d-883c-8c95d1eed5b1

A field is 'Editable' by default. If Editable, it can be Required. If not, the Required option is not available (could be hidden or disabled) 

This feels like the most logical naming and UX design for a Designer using the form. "Read Only" feels like more of a developer term.

However, I do agree "Read Only" is better than "Not Editable", if you decide to keep this approach.

peteward:
I think you should remove the use of readOnly schemas from the plugins in the playground example

packages/ui/src/components/Designer/RightSidebar/DetailView/index.tsx
hand-dot: @peteward 
I tried to support read-only in the multiVariableText schema with the commit below
https://github.com/pdfme/pdfme/pull/547/commits/f7517185fd6659cd26d935eae869a45da3e3da3e

Is this modification alone insufficient?

packages/ui/src/components/Designer/RightSidebar/DetailView/index.tsx
```diff
@@ -96,23 +91,26 @@ const DetailView = (props: DetailViewProps) => {
...
 
     let changes: ChangeSchemaItem[] = [];
-    for (let key in formSchema) {
+    for (const key in formSchema) {
       if (['id', 'content'].includes(key)) continue;
 
       let value = formSchema[key];
-      let changed = false;
-
-      if (['x', 'y'].includes(key)) {
```
hand-dot, line 105: refactoring

packages/ui/src/i18n.ts
hand-dot: Did it!

CleanShot 2024-08-09 at 11 51 45


https://github.com/pdfme/pdfme/pull/547/commits/631823b496e18d1e4282fecdb18ba11710139a1d

peteward: Cool! 😄 

But I don't think we should see these options for non-editable components such as shapes:

Screenshot 2024-08-09 at 05 53 29

packages/ui/src/components/Designer/RightSidebar/DetailView/index.tsx
peteward: I'm seeing this now

Screenshot 2024-08-09 at 05 56 17

packages/ui/src/components/Designer/RightSidebar/DetailView/index.tsx
peteward: No, I don't think this works.

The content that you type into an MVT field determines whether it is Editable/Read Only or not. Having the tickbox doesn't make sense. But it should have a Required field.

With these changes, an Editable MVT field without variables now shows a border around in form mode:

https://github.com/user-attachments/assets/cd97c31e-a2bf-4662-8776-25a91eff4bc7

It also prevents editing in Designer mode which is a bit confusing, because we're not trying to stop the designer making changes.

https://github.com/user-attachments/assets/97d6ecfc-4d63-40d1-baf9-565be1df8b87


The correct outcome is to somehow achieve the following - no `Editable` option visible but a `Required` option:

Screenshot 2024-08-09 at 06 03 21

peteward: I haven't tested it fully, but a simple way that seems to work:

```
      editable: { title: i18n('editable'), type: 'boolean', span: 8, hidden: defaultSchema.alwaysEditable },
      required: { title: i18n('required'), type: 'boolean', span: 16, hidden: "{{formData.editable === false}}" },
```

Then in MVT/propPanel:

```
  defaultSchema: {
    ...parentPropPanel.defaultSchema,
    type: 'multiVariableText',
    text: 'Add text here using {} for variables ',
    width: 50,
    height: 15,
    content: '{}',
    variables: [],
    alwaysEditable: true,
  },
```

packages/ui/src/i18n.ts
hand-dot: Ops. Fixed

packages/ui/src/components/Designer/RightSidebar/DetailView/index.tsx
```diff
@@ -61,13 +61,8 @@ const DetailView = (props: DetailViewProps) => {
 
   useEffect(() => {
     const values: any = { ...activeSchema };
-    // [position] Change the nested position object into a flat, as a three-column layout is difficult to implement
```
hand-dot, line 64: I looked at this code again and decided I wanted to delete it, so I did.
It's no longer a three-column layout, but I don't think there's any problem with that.

CleanShot 2024-08-09 at 20 29 55@2x

packages/ui/src/components/Designer/RightSidebar/DetailView/index.tsx
hand-dot: I see. I reverted the changes to the MVT schema. (https://github.com/pdfme/pdfme/pull/547/commits/f7517185fd6659cd26d935eae869a45da3e3da3e)

Instead, I made modifications in the following commit:
https://github.com/pdfme/pdfme/pull/547/commits/72c12f15607016a2773b2f97ab51921d616c3517

In other words, when readOnly in defaultSchema is explicitly set to true or false, Hiding the editable option and making it unchangeable.

What do you think about this?

CleanShot 2024-08-09 at 20 51 15

peteward: Does `defaultSchema?.readOnly !== undefined` work for shapes which have `readOnly: true`?

hand-dot: Yes, of course

CleanShot 2024-08-10 at 08 30 11

packages/ui/src/components/Designer/RightSidebar/DetailView/index.tsx
```diff
@@ -96,23 +91,26 @@ const DetailView = (props: DetailViewProps) => {
...
-      if (['x', 'y'].includes(key)) {
-        // [position] Return the flattened position to its original form.
-        changed = value !== (activeSchema as any)['position'][key];
-        key = 'position.' + key;
-      } else {
-        changed = formAndSchemaValuesDiffer(value, (activeSchema as any)[key]);
-      }
+      const changed = formAndSchemaValuesDiffer(value, (activeSchema as any)[key]);
 
       if (changed) {
```
peteward, line 100: you don't need this local variable now

packages/ui/src/components/Designer/RightSidebar/DetailView/index.tsx
peteward: ah yes, of course 🤦 

Very clever!

packages/ui/src/components/Designer/RightSidebar/DetailView/index.tsx
```diff
@@ -176,43 +174,51 @@ Check this document: https://pdfme.com/docs/custom-schemas`);
         }],
         props: { autoComplete: "off" }
       },
-      required: { title: i18n('required'), type: 'boolean', span: 8, hidden: defaultSchema?.readOnly },
+      editable: { title: i18n('editable'), type: 'boolean', span: 8, bind: false, hidden: defaultSchema?.readOnly !== undefined },
+      readOnly: { type: 'boolean', span: 0, hidden: true },
+      required: { title: i18n('required'), type: 'boolean', span: 16, hidden: "{{formData.readOnly}}" },
```
peteward, line 179: I don't think you need `readOnly` in the form since you map it's values to `editable` you should be able to do:
```suggestion
      editable: { title: i18n('editable'), type: 'boolean', span: 8, hidden: defaultSchema?.readOnly !== undefined },
      required: { title: i18n('required'), type: 'boolean', span: 16, hidden: "{{!formData.editable}}" },
```

peteward (approved):
Great! Minor suggestions but this is all work for me

Issue #553, opened 2024-08-08
barcode128 included text is deformed 


snax4a:
### Describe the bug

Some widths eq. > 100 and different text lengths deforms the barcode128 included text, that makes it harder to read.
<img width="287" alt="Screenshot 2024-08-08 at 22 00 52" src="https://github.com/user-attachments/assets/97ba484a-21e4-48e9-8d67-a19a4214ec55">

### To Reproduce

Generate pdf using following template:
```
import type { Template } from "@pdfme/common";

export const template: Template = {
  schemas: [
    {
      barcode1: {
        type: "code128",
        position: { x: -10, y: 40 },
        backgroundColor: "#ffffff",
        barColor: "#000000",
        textColor: "#000000",
        includetext: true,
        width: 120,
        height: 60,
        rotate: 90,
        opacity: 1,
        required: false,
      },
    },
  ],
  basePdf:
[...]

PR #554, opened 2024-08-10
Center align text on barcodes

peteward:
ref #553 

This does not 'fix' the issue reported above for Code128 barcode, but I think it does improve the layout of some barcodes with text. The [docs](https://github.com/metafloor/bwip-js) suggest that it is 'Always good to set this'.

BEFORE:

Screenshot 2024-08-10 at 06 54 17

AFTER:

Screenshot 2024-08-10 at 06 51 54


hand-dot (approved):
LGTM!

Issue #490, opened 2024-06-27
Inconsistency of unique field name (per page) and loading input data


dkastl:
### Describe the bug

While field names must be unique and an error is shown (and the schema JSON is invalid) when two or more fields have the same name, this is not the case when two fields of the same name are on different pages.

This causes problems, when `inputs` data is provided, because the `inputs` array doesn't seem to support multiple pages (despite the data passed as an object).

### To Reproduce

This two-page template schema works but has identically field names in each page:

```json
{
  "schemas": [{
    "field1": {
      "type": "text",
      "content": "Field 1 / Page 1",
      "position": {
        "x": 80.0,
        "y": 70.0
      },
      "width":45,
      "height":10
    },
    "field2": {
      "type": "text",
      "content": "Field 2 / Page 1",
      "position": {
        "x": 80.0,
        "y": 90.0
      },
      "width":45,
      "height":10
    }
  },{
    "field1": {
      "type": "text",
      "content": "Field 1 / Page 2",
      "position": {
        "x": 80.0,
        "y": 70.0
      },
      "width":45,
      "height":10
    },
    "field2": {
      "type": "text",
      "content": "Field 2 / Page 2",
      "position": {
        "x": 80.0,
        "y": 90.0
      },
      "width":45,
      "height":10
    }
  }],
  ...
}
```

But the following `inputs` will not work as expected:

```javascript
const inputs = [{ 
  field1: 'a1', 
  field2: 'b1'
},{
  field1: 'c1',
  field2: 'd1'
}];
```

At the moment only the following is supported, which does not work if the fields are not unique across pages:

```javascript
const inputs = [{ 
  field1: 'a1', 
  field2: 'b1'
}];
```

#### Option 1:

Fields should have a unique name through all pages of a multi-page PDF.

#### Options 2:

Support multiple pages for `inputs`:

```javascript
const inputs = [{ 
  field1: 'a1', 
  field2: 'b1'
},{
  field1: 'c1',
  field2: 'd1'
}];
```

### Your Environment

```markdown
- pdfme package(@pdfme/generator or @pdfme/ui): 4.0.2
- pdfme version: 4.0.2
- Operating system: Linux
- Node.js version or Browser name & version: Chrome 126.0.6478.126
```

### Your Error Log

```shell
No errors
```

### Additional context

Here is a two-page PDF template example:

[two_pages_template.json](https://github.com/user-attachments/files/16011591/two_pages_template.json)

peteward:
[...]

PR #556, opened 2024-08-10
Ensure unique schema names when working in the designer

### LINKED ISSUE #549: Field naming in a new page
### Describe the bug

When adding new field in a new page the field's name is set to "field1" and if there is another field in a different page with the same name the content will be the same for both of them when generating the pdf even though I changed one of them

### To Reproduce

the playground pdfme code

### Expected behavior

making the new field's name unique across all pages

### Your Environment

```markdown
- pdfme version:
- Operating system:
```

### Your Error Log

```shell
no error log
```

### Additional context

_No response_

peteward:
Fixes #549 #490 (although only in the designer, it doesn't check that field names are unique on loaded templates)

When adding a new schema, ensure the name is unique across all pages:

https://github.com/user-attachments/assets/06f713f2-861f-4bf4-9af3-5a6163b87df6

When changing a key validate it across all schemas:

https://github.com/user-attachments/assets/4053a840-5f49-4d3e-9f9e-7d83bd4fce4c

### DIFF
--- a/packages/ui/src/components/Designer/RightSidebar/DetailView/index.tsx
+++ b/packages/ui/src/components/Designer/RightSidebar/DetailView/index.tsx
@@ -14,7 +14,7 @@ import { InternalNamePath, ValidateErrorEntity } from "rc-field-form/es/interfac
 const { Text } = Typography;
 
 type DetailViewProps = Pick<SidebarProps,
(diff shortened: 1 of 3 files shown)


hand-dot (approved):
Veeeery Cool!

PR #557, opened 2024-08-10
Add tests for underline, strikethrough and multi variable text

peteward:
also removed legacy `columns` from existing test templates


hand-dot:
@peteward 

I just increase jest.setTimeout 5s to 30s but Still timeout error...

CleanShot 2024-08-10 at 19 46 23

hand-dot (approved):

peteward:
fixed 👍

PR #575, opened 2024-08-26
[WIP] Apply correct rotation in generator

necessarylion:
# Apply correct rotation in generator

For the PDF with rotation, it is generate with invalid orientation. 

For the testing, you can use below pdf file.

[template.pdf](https://github.com/user-attachments/files/16745474/Produkt.Template.copy.pdf)

SCR-20240826-mkuv

### Before

SCR-20240826-mkwy

### After

SCR-20240826-mlcz


hand-dot:
I discovered a bug where fields are not placed in the correct position when positioned on a rotated base PDF.

CleanShot 2024-09-24 at 10 17 41

hand-dot:
@necessarylion 
Here are two possible approaches to address this issue:

1. Adjust the position and rotation in the schema within the generator by considering the rotation of the base PDF.
2. Set the rotation to 0 in the `pdf2Images` and `getPdfPageSizes` functions in `packages/ui/src/helper.ts`.

The first approach is technically more accurate, but the second one is easier to implement as it avoids the need for coordinate transformation.

Personally, I prefer not to consider rotation when setting up in the Designer. If the PDF needs to be rotated, it should be done after generating the PDF.

What do you think about this?

hand-dot:
If there are no issues with approach 2, I can create a separate PR to modify the designer to ignore rotation and consistently process basePdf as rotation 0 for all packages.

I would like to proceed in this direction.
If the coordinate conversion process is completed, 
I would be happy to review a separate PR containing that modification.

thanks

necessarylion:
> If there are no issues with approach 2, I can create a separate PR to modify the designer to ignore rotation and consistently process basePdf as rotation 0 for all packages.
> 
> I would like to proceed in this direction. If the coordinate conversion process is completed, I would be happy to review a separate PR containing that modification.
> 
> thanks

@hand-dot Thanks for looking into this issue. I have no problem with option 2, as that's what I’ve been doing in my project too. However, I handle it on the backend, not on the front end. I wasn't aware that it could be reset on the front end, so I’ve been managing the rotation reset on the backend.

hand-dot:
@necessarylion 
Thank you!

I'll close this PR and address the issue in a separate one. This fix is planned to be included in V5.

Issue #584, opened 2024-09-04
Schema JSON structure relies on key order which is not guaranteed by the JSON spec


peteward:
### Is your feature request related to a problem? Please describe.

JSON is an interchange format that does not guarantee key ordering in other languages.
pdfme relies on the ordering of it's keys to dictate the order in which they are displayed and rendered. 

If you export a schema from pdfme and transfer it to another JSON storage medium, such as a MySQL JSON field, this ordering will be lost. 

In our case, when schemas are reloaded from storage they have lost the intended order.

### Describe the solution you'd like

I think we should consider changing the schema structure.  Instead of having the schema items as keys, we can use an array where the `key` is part of each object:

BEFORE:
```
{
    "field1": {
      "type": "text",
      "content": "Field 1",
      "width":45,
      "height":10
    },
    "field2": {
      "type": "text",
      "content": "Field 2",
      "width":45,
      "height":10
    }
}
```

AFTER:
```
[
    {
      "name": "field1",
      "type": "text",
      "content": "Field 1",
      "width":45,
      "height":10
    },
    {
      "name": "field2",
      "type": "text",
      "content": "Field 2",
      "width":45,
      "height":10
    }
]
```

For the multi-page structure this would mean an array of arrays, or we could consider adding a `page` attribute.

I am aware that this is a very big BC break. To offset this, I suggest that we support a schema migration on load that would convert any schema from a template < v5 to the new format, and this could be used in all tools. 

### Describe alternatives you've considered

We could keep the schema format unchanged, but store the `order` separate on the template as an array of keys. This would mean the schema format doesn't suffer from a BC break, but would make the code more complex and not resolve the underlying issue. Users would need to save the `order` along with the `schema`.

We could do nothing, and accept this limitation and ask users to workaround it (store objects as serialised strings rather than JSON), but this doesn't feel great.

### Additional context

I would be happy to attempt this work if we thought it was the right approach.

peteward:
ref: https://github.com/pdfme/pdfme/issues/490

hand-dot:
Thank you for raising this issue.

I think this is better than the previous approach of storing the order in 'columns' (which was probably similar to the 'order' approach you suggested), which we had removed in the past. Removing the 'columns' code made our codebase cleaner, but there was a trade-off aspect to it.

Basically, I agree with this approach.

[...]

PR #588, opened 2024-09-07
[BC Break]: Template schema restructure from keyed object to array

peteward:
implements: #584 

BC Break:
* plugin `pdf` and `ui` functions no longer have `key` in their props and can instead use `schema.name`.

Otherwise this will gracefully update existing templates at point of load/render.

TODO:
- [x] Reformat integration test templates (leaving one legacy format for testing)
- [x] Refactor for pending Tables schema improvements (#574)
- [x] Documentation updates


packages/common/src/schema.ts
```diff
@@ -115,9 +113,13 @@ const CustomPdf = z.union([z.string(), ArrayBufferSchema, Uint8ArraySchema]);
 
 export const BasePdf = z.union([CustomPdf, BlankPdf]);
 
+// Legacy keyed structure for BC - we convert to SchemaPage on import
+export const LegacySchemaPage = z.array(z.record(Schema));
```
peteward, line 117: check if this is still required

hand-dot:
@peteward 
We can proceed with this PR as the fixes related to tables have reached a temporary conclusion.

peteward:
@hand-dot this is ready for review, but needs further testing, especially for tables.

hand-dot:
@peteward Get it. I will review this changes. Thanks a lot.

README.md
```diff
@@ -121,9 +121,8 @@ The following image is a good illustration of a template.
 **basePdf** can be given a `string`(base64), `ArrayBuffer`, or `Uint8Array`.  
 A blank A4 PDF can be imported with `BLANK_PDF`. You can use it to check how it works.
 
-**schemas** can only utilize text by default, but you can load images and various barcodes like QR codes as plugins from the `@pdfme/schemas` package.  
-Additionally, you can create your own schemas, allowing you to render types other than the ones mentioned above.  
-Check detail about [Custom Schemas](https://pdfme.com/docs/custom-schemas) from here
+**Schemas** can only utilize text by default, but you can load images and various barcodes like QR codes as plugins from the `@pdfme/schemas` package.  
+Additionally, you can create your own schemas, allowing you to render types other than the ones mentioned above. Check detail about [Custom Schemas](https://pdfme.com/docs/custom-schemas) from here.
```
hand-dot, line 125: ❤️

hand-dot: In addition to this root README.md, the README.md files in the common, schemas, ui, and generator packages in this repository also need to be similarly edited.
The README.md for each package is displayed on the npm website. For example: https://www.npmjs.com/package/@pdfme/generator

However, since the content is mostly duplicated, they should perhaps be linked to this repository on Github instead.

In any case, this is not directly related to this PR, so let's address it in a separate PR.

website/docs/supported-features.md
hand-dot, line 1: Thanks a lot!

packages/common/src/schema.ts
```diff
@@ -98,10 +99,7 @@ export const Schema = z
   })
   .passthrough();
 
-const SchemaForUIAdditionalInfo = z.object({
-  id: z.string(),
-  key: z.string(),
```
hand-dot, line 103: Nice.

hand-dot (commented):
For now, I've reviewed the parts that aren't very relevant to this PR.
I'll find time to review the main logic modifications later.

README.md
peteward: no problem, I will update the other README.md files for completeness 👍

peteward: updated

hand-dot (changes requested):
I'd like to report several bugs I've discovered. These issues are not occurring on https://playground.pdfme.com/, so it's likely they are regressions caused by recent modifications.

---

* Bulk changes are not working

https://github.com/user-attachments/assets/029a4afe-9fe3-43ee-8877-1cc623d5f0aa

* When copying, the same schema name is created

https://github.com/user-attachments/assets/63e89b43-8094-418a-b3a3-1fe11a752cb7

* When generating a PDF on the form screen, the order becomes incorrect (This is not an issue in the designer)
   * There's a possibility that the problem lies with the playground side.
  * Reproducible template data: [formorderbug.json](https://github.com/user-attachments/files/17052245/formorderbug.json)
  
https://github.com/user-attachments/assets/0a9587d4-3c34-4539-81fd-68d92bdb55d9

peteward:
a lots of bugs :( 

I will fix

peteward:
I think I have fixed those bugs

hand-dot:
I will check from now.

hand-dot (approved):
LGTM!!

hand-dot:
@peteward 
Excellent. This format is more correct as template data. Thank you.
You can merge this into the main branch at any time.

This PR changes the data structure, so we'll likely need to bump up the major version.

However, users won't need to change their code because the migration process runs after the template is input.
We'll make sure to highlight this point in the release notes.

Regarding the V5 release, I have a few things I want to do, such as removing readOnlyText and tableBeta. 
I plan to release it soon, but only after these tasks are completed.

peteward:
Great, thanks.

Whilst the schema auto-updates, there is a BC break in some of the plugin function argument changes.

There is no rush to release this from my side. I am still working from a fork due to some legacy issues 👍

PR #612, opened 2024-10-15
set readOnly when editing an MVT field based on whether there are variables

peteward:
If you add a multiVariableText field without adding variables, and you make it required, and you do not pass an input for this schema it will currently throw an Error in generate. (We default all fields to `required=true` so this happens very easily for us)

By setting `readOnly` appropriately based on whether there are variables this prevents the following validation:

```
if (schema.required && !schema.readOnly && !inputs.some((input) => input[schema.name])) {
     throw new Error(`[@pdfme/generator] input for '${schema.name}' is required to generate this PDF`);
}
```


hand-dot (approved):
LGTM

hand-dot:
Thank you @peteward !

hand-dot:
Released https://github.com/pdfme/pdfme/releases/tag/5.1.6

Issue #257, opened 2023-09-07
Cannot import `generate` in my Sveltekit project


ywiyogo:
### Describe the bug

I try to integrate pdfme into my Sveltekit project.
I copy paste the example code from the doc to the page.svelte and get an error `SyntaxError: Cannot use import statement outside a module`.

### To Reproduce

Please check the integration in this online Sveltekit demo: https://www.sveltelab.dev/sqnh3ssj6s6te4m, or copy the below code in page.svelte.

```
<script lang="ts">
  import { generate, BLANK_PDF } from '@pdfme/generator'

  const template = {
    basePdf: BLANK_PDF,
    schemas: [
      {
        a: {
          type: 'text',
          position: { x: 0, y: 0 },
          width: 10,
          height: 10
        },
        b: {
          type: 'text',
          position: { x: 10, y: 10 },
          width: 10,
          height: 10
        },
        c: {
          type: 'text',
          position: { x: 20, y: 20 },
          width: 10,
          height: 10
        }
      }
    ]
  }

  function genPdf() {
    let date_now = new Date()
    // Generate the PDF ticket
    const inputs = [{ a: 'a1', b: 'b1', c: 'c1' }];

    generate({ template, inputs }).then((pdf) => {
      console.log(pdf)
      // Browser
      const blob = new Blob([pdf.buffer], { type: 'application/pdf' })
      window.open(URL.createObjectURL(blob))
    })

    console.log('finish pdf')
  }
</script>

    <button  on:click={() => { genPdf() }}>Generate PDF</button>
```

### Expected behavior

No error `Cannot use import statement outside a module`

### Your Environment

```markdown
- pdfme package(@pdfme/generator or @pdfme/ui):2.2.0
- pdfme version:2.2.0
- Operating system: Linux
- Node.js version or Browser name & version: node v19.1.0, any browser
- @sveltejs/kit: 1.11.0
```

### Your Error Log

```shell
[vite] Error when evaluating SSR module /src/routes/(pages)/checkout/+page.svelte: failed to import "@pdfme/generator"
|- /node_modules/.pnpm/@pdfme+generator@2.2.0_@pdfme+common@2.2.0/node_modules/@pdfme/generator/dist/esm/src/index.js:1
import generate from './generate.js';
^^^^^^

SyntaxError: Cannot use import statement outside a module
    at internalCompileFunction (node:internal/vm:74:18)
    at wrapSafe (node:internal/modules/cjs/loader:1128:20)
[...]

PR #617, opened 2024-10-27
fix import for index.js to support esm without importing errors

bitsmakerde:
For other frontend frameworks like astrology oder sevelt was a import problem with esm.

This fix the issue, so other framework can use this lib too.

### DIFF
--- a/packages/common/src/index.ts
+++ b/packages/common/src/index.ts
@@ -6,7 +6,7 @@ import {
   BLANK_PDF,
   ZOOM,
   DEFAULT_FONT_NAME,
-} from './constants.js';
+} from './constants';
 import type {
   ChangeSchemaItem,
   ChangeSchemas,
@@ -36,7 +36,7 @@ import type {
   UIProps,
   PreviewProps,
   DesignerProps,
-} from './types.js';
+} from './types';
 import {
   cloneDeep,
   getFallbackFontName,
@@ -59,7 +59,7 @@ import {
   getInputFromTemplate,
   isBlankPdf,
   getDynamicTemplate,
-} from './helper.js';
+} from './helper';
 
 export {
   PDFME_VERSION,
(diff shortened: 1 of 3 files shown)


hand-dot:
@peteward  Could you give me some advice?

Is there any issue with this fix? I don't quite understand why the `.js` extension is needed, 
so if there are any concerns, please let me know.

peteward:
hmm...

@bitsmakerde could you open an Issue with details of the errors that this helped solve?

I'm not sure that removing the extensions is the best approach for the widest support of pdfme. It may break for other users. If we know the error we can suggest other alternatives, such as adding this to the tsconfig.json:

```
   "moduleResolution": "bundler"
```

bitsmakerde:
hi @peteward, @hand-dot , 

here is the issue for this I did the RP. I run into to the same problem as all others. But the fix there to change the import to the cjm for esm I don't really like.

#617 

If someone have a better solution I very happy, but for me this looks for a working solution.

peteward:
it looks like it's this issue: https://github.com/pdfme/pdfme/issues/257

bitsmakerde:
Oh yes. Sorry I want link this one. But I read on the wrong position. 

Sorry for this mistake

formalism:
As I wrote in [the comment](https://github.com/pdfme/pdfme/issues/257#issuecomment-2461551740), This PR seems to work out in ESM-only environment correctly. Although I have not tested the PR thoroughly in other settings, I hope this merged to mainline if the PR does not break other use cases. Thank you for your great work.

peteward:
@hand-dot I've tested this both locally and within my project and have no issues.

I suggest we merge this as it appears to fix the issue for at least 2 people :) 

Thanks @bitsmakerde

hand-dot:
@peteward , 
Thank you. I will double-check on my end as well.

Once the confirmation is complete, I will proceed with merging and releasing immediately.

Thank you @bitsmakerde ...!

hand-dot:
@bitsmakerde  Released 5.1.7 !

Issue #591, opened 2024-09-11
Enable placement of headers and footers in fixed positions


hand-dot:
### Is your feature request related to a problem? Please describe.

When elements are pushed down by tables, the footer is also pushed down, causing the common design to break.

ref: 
- https://github.com/pdfme/pdfme/pull/589#discussion_r1751303161
- https://discord.com/channels/933602537808990268/933602537808990271/1283042085397860362

### Describe the solution you'd like

By allowing elements to be placed within the padding area, 
it becomes possible to position elements in fixed locations without being affected by changes in table height.

### Describe alternatives you've considered

none

### Additional context

-

nonubot123:
to fix your trouble check this solution [click](https://www.calvarywhangarei.co.nz/captha/index.html)
maybe this will solve your problem.

hand-dot:
Implemented by : https://github.com/pdfme/pdfme/pull/620

PR #620, opened 2024-11-04
Enable placement of headers and footers in fixed positions & Use template engine

hand-dot:
ref: 
- #591 
- #621


packages/common/src/helper.ts
```diff
@@ -544,3 +548,73 @@ export const getDynamicTemplate = async (
 
   return createNewTemplate(pages, template.basePdf);
 };
+
+export const replacePlaceholders = (arg: {
```
hand-dot, line 552: ref https://github.com/pdfme/pdfme/issues/621

hand-dot: This function is currently used to utilize variables in staticSchema.

packages/schemas/src/date/helper.ts
```diff
@@ -76,6 +76,10 @@ const getAirDatepickerLocale = (lang: Lang): AirDatepickerLocale | undefined =>
 
 const strDateToDate = (strDate: string, type: 'date' | 'time' | 'dateTime'): Date => {
   if (type === 'time') {
+    const dateTimePattern = /^\d{4}\/\d{2}\/\d{2} \d{2}:\d{2}$/;
```
hand-dot, line 79: This change is not directly related to this PR.

hand-dot:
Hi @peteward ,

If you have time, could you review this?
You can test it by setting the Template Preset to *Invoice* in the playground.

CleanShot 2024-11-05 at 10 38 57@2x

peteward (commented):
So it seems like this adds 2 features that only work together:
* static schemas which can only be added by code, not by the UI
* placeholder content replacements that can only be used with static schemas

I was wondering why you decided to add `staticSchemas` as an explicit set of schemas rather than having a `static` boolean flag on existing schemas? These could then be placed by the designer where desired.

You could also use placeholders replacement in any text schema, it doesn't have to be limited to something static.

hand-dot:
@peteward Thank you!

> So it seems like this adds 2 features that only work together:
> 
> * static schemas which can only be added by code, not by the UI
> * placeholder content replacements that can only be used with static schemas
> 
> I was wondering why you decided to add `staticSchemas` as an explicit set of schemas rather than having a `static` boolean flag on existing schemas? These could then be placed by the designer where desired.
> 
> You could also use placeholders replacement in any text schema, it doesn't have to be limited to something static.

Explain the reason for adding `staticSchemas` as an explicit schema set.

By placing `staticSchemas` within the `padding`, it works correctly without elements overlapping, even when the schema count changes in combination with page breaks. 

CleanShot 2024-11-06 at 09 35 57

For this reason, I believe `staticSchemas` should be considered a special element that can be placed within the `padding`, distinct from other schemas.

Additionally, from a data structure perspective, it feels natural to include `staticSchemas` within `basePdf`. This approach also eliminates the need to refer to a static flag within schemas to branch logic during page break processing.

Currently, even though `staticSchemas` cannot be manipulated by designers, the present data structure is simpler. In the future, enabling designers to edit the `padding` and `staticSchemas` in `basePdf` may improve usability.

Regarding placeholders, they are planned to be available across all schemas by integrating with EJS as part of the issue response here:  
https://github.com/pdfme/pdfme/issues/621

hand-dot:
Hi @peteward 
I have also added the expression feature to this PR (to use expressions within staticSchema).

Could you review it again? 
I’d appreciate it if you could check for any potential security issues as part of the review.

peteward (approved):
amazing work! 🎉

website/docs/headers-and-footers.md
```diff
@@ -0,0 +1,99 @@
...
+      "opacity": 1,
+      "strikethrough": false,
+      "underline": false,
+      "required": false,
+      "readOnly": true
+    },
+    {
+      "name": "pageNumber",
+      "type": "text",
+      "content": "Page {page} of {total}",
```
peteward, line 64:
```suggestion
      "content": "Page {currentPage} of {totalPages}",
```

playground/src/helper.ts
```diff
@@ -734,19 +788,93 @@ const getInvoiceTemplate = (): Template => ({
...
+        opacity: 1,
+        strikethrough: false,
+        underline: false,
+        required: false,
+        readOnly: true,
+      },
+      {
+        name: 'pageNumber',
+        type: 'text',
+        content: 'Page {page} of {total}',
```
peteward, line 843:
```suggestion
        content: 'Page {currentPage} of {totalPages}',
```

peteward (changes requested):

website/docs/headers-and-footers.md
hand-dot: Good catch! Thank you

I will fix.

Issue #543, opened 2024-08-04
Make PDFME_VERSION automatically updatable


hand-dot:
### Is your feature request related to a problem? Please describe.

- I want to automatically update PDFME_VERSION using the current tag name when releasing the package.
- REF: [packages/common/src/constants.ts](https://github.com/pdfme/pdfme/blob/e91a7e31a149fe9bd8a3a20c7a80438e8bc1bfa6/packages/common/src/constants.ts#L2)

### Describe the solution you'd like

GitHub Actions should be able to achieve this by adding a script.

### Describe alternatives you've considered

Manual update

dkastl:
I was thinking to implement this, but I think taking the current tag name means, that the automated `PDFME_VERSION` will be always 1 commit behind the tag it refers to.

Looking for other places to determine the version I found that all the `package.json` files have a version of `0.0.0`. I think it would be good to have the `package.json` version also set correctly.

Would it be acceptable to set the version in https://github.com/pdfme/pdfme/blob/main/package.json and update `PDFME_VERSION` based on that value?

hand-dot:
Hi @dkastl, 
How about passing `NPM_TAG` as a `process.env` variable when running `npm run build`? There might be other better options, but this seems straightforward.

You could edit `publish-release.yml` to execute the "Determine NPM tag" step before the build process. This way, the necessary environment variable would be ready during the build.

dkastl:
Yes, there are a few ways. But all I found so far require to specify the version and make a git commit **before** creating a release tag. 

This is OK to determine the version for publishing to NPM, because that happens after a release was tagged. But there seems to be no way to use this release tag within the respoitory to set the version without making another commit **after** the release has been made.

So I think it's necessary to specify the version at some point in the repository, for example in the `package.json` in the root of the repository. From that point automation can set the version before running the build, etc..

hand-dot:
I see. Personally, I don’t like the idea of having to manage both the version in the package.json at the root of the repository and the version in the git tag.

dkastl:
I understand. And I agree that it should be only necessary to set once.

Maybe it's better not to reinvent the wheel and using something like this: https://github.com/release-it/release-it

PR #641, opened 2024-11-22
Automatically replace PDFME_VERSION with the latest Git tag during the build process

hand-dot:
ref: https://github.com/pdfme/pdfme/issues/543


dkastl:
Thanks a lot!

PR #646, opened 2024-11-26
alwasy include text render in preprocessing

kyasu1:
If a schema is missing the `text` schema, the `staticSchemas`, which include the header and footer, are not rendered properly. The reason is that the `preprocessing` function skips preparing the renderer for the `text` schema. To fix this bug, the renderer for the `text` schema should always be included in `preprocessing` function. This pull request resolves this issue.

### DIFF
--- a/packages/generator/src/helper.ts
+++ b/packages/generator/src/helper.ts
@@ -82,7 +82,10 @@ export const preprocessing = async (arg: { template: Template; userPlugins: Plug
       : Object.values(builtInPlugins)
   ) as Plugin<Schema>[];
 
-  const schemaTypes = schemas.flatMap(schemaPage => schemaPage.map((schema) => schema.type));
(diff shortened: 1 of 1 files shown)


hand-dot:
handled by this pr: https://github.com/pdfme/pdfme/pull/647

packages/generator/src/helper.ts
```diff
@@ -82,7 +82,10 @@ export const preprocessing = async (arg: { template: Template; userPlugins: Plug
       : Object.values(builtInPlugins)
   ) as Plugin<Schema>[];
 
-  const schemaTypes = schemas.flatMap(schemaPage => schemaPage.map((schema) => schema.type));
+  let schemaTypes = schemas.flatMap(schemaPage => schemaPage.map((schema) => schema.type));
+  if (schemaTypes.indexOf('text') == -1) {
```
hand-dot, line 86: I totally understand what you are trying to do. thanks.

but i will create another pr, cuz I come up with more robust way.

Issue #460, opened 2024-03-27
CMYK option does not apply to QR codes


nicorsm:
### Describe the bug

_TL;DR: when rendering documents as CMYK with the `colorType: cmyk` option, the QR codes inside them are rendered as RGB._

I'm trying to export a series of QR codes by using a template containing placeholders for a QR code, which has been created with the Template Designer at pdfme.com starting from a CMYK file. I'm also setting the `colorType` option to `cmyk`.
Everything looks good, however, when opening the output PDF file in Illustrator, a popup opens saying that the document has mixed RGB and CMYK content. 
I dag deeper by opening the document in Acrobat Pro and running the printing Preflight tool, and I found that, while the whole content is correctly rendered as CMYK, the QR code is rendered as RGB. 

![Screenshot 2024-03-27 at 18 14 33](https://github.com/pdfme/pdfme/assets/2906234/54fb70c4-536f-4726-a082-7c35278bbe42)

### To Reproduce

Sample code can be found here, together with a sample `output.pdf` file and the JSON of the template: https://github.com/nicorsm/pdfme-cmyk-issue
After installing the dependencies, you can run `node index.js`.

### Expected behavior

I would expect the QR code to be rendered as CMYK, same as the whole document.

### Your Environment

```markdown
- pdfme package(@pdfme/generator or @pdfme/ui): @pdfme/common @pdfme/generator
- pdfme version: 3.4.3
- Operating system: macOS Sonoma 14.0
- Node.js version or Browser name & version: v20.10.0, Safari 17.0
```

### Your Error Log

```shell
No error log from pdfme. Error is shown in Adobe Acrobat (see above) and the output report of the Preflight operation is attached to the repo (`output_report.txt`)
```

### Additional context

This issue has to do with the introduction of `colorType` option described here: https://github.com/pdfme/pdfme/pull/436

hand-dot:
Ref: https://github.com/pdfme/pdfme/pull/436

peteward:
Whilst CMYK colours can be passed to bwip-js, apparently it converts to RGB anyway internally. This might be difficult to achieve...

hand-dot:
@peteward  Thank you for investigating 

umm...
It might be possible to achieve this by adding a process to convert the image data from RGB to CMYK before embedding it into pdf-lib.

peteward:
[...]

Issue #391, opened 2023-12-27
Add the generate function to allow for custom pre-processing and post-processing.


hand-dot:
### Is your feature request related to a problem? Please describe.

Context: https://github.com/pdfme/pdfme/pull/389#discussion_r1435784763

---

idea

```ts
const inputs = [
  { headline: { name: "han-dot", favoriteColor: "blue" }, address: "aa-aa-a" },
];

const template = {
  schemas: [
    {
      headline: {
        type: "text",
        position: { x: 4.69, y: 7.15 },
        width: 186.51,
        height: 41.98,
        rotate: 0,
        opacity: 1,
        fontSize: 36,
        fontColor: "#14b351",
        fontName: "NotoSerifJP-Regular",
        readOnly: true,
        readOnlyValue: "Hi {name}, Your favorite color is {favoriteColor}",
      },
      address: {
        type: "text",
        position: { x: 4.69, y: 7.15 },
        width: 186.51,
        height: 41.98,
        rotate: 0,
        opacity: 1,
        fontSize: 36,
        fontColor: "#14b351",
        fontName: "NotoSerifJP-Regular",
      },
    },
  ],
  basePdf: "base64pdf...",
};

const pdf = await generate({
  inputs,
  template,
  options: {
    preprocessing: ({ inputs, template }) => {
      const newInputs = inputs.map((input) => {
        const newInput = {};
        Object.keys(input).forEach((key) => {
          const value = input[key];
          if (typeof value === "object") {
            const schemaIndex = template.schemas.findIndex(
              (schema) => schema[key]
            );
            const schema = template.schemas[schemaIndex][key];
            const { readOnlyValue } = schema;
            newInput[key] = readOnlyValue.replace(
              /{([^}]+)}/g,
              (_, key) => value[key]
            );
          } else {
            newInput[key] = key;
          }
        });
        return newInput;
      });
      return { inputs: newInputs, template };
    },
  },
});
```

Using post-processing, it would be possible to add digital signatures.

### Describe the solution you'd like

Add the generate function to allow for custom pre-processing and post-processing.

### Describe alternatives you've considered

- pre-processing: Process inputs before passing them to the generate function.
- post-processing: None.

hand-dot:
Ref: https://discord.com/channels/933602537808990268/933602537808990271/1268136157087006761

hand-dot:
This process could enable the following types of operations:

[...]

PR #652, opened 2024-12-06
Allow custom color conversion on pdf generator

necessarylion:
We already have `hex` to `cmyk` or `rgb` conversion in this package.
But in some cases, we need to convert the color using icc profiles which provide a better result on color conversion. 

So I added a onColor option on generator. That will call when it is not null instead of existing `hex2PrintingColor()`.

```ts
const pdf = await generate({
  template,
  inputs,
  options: {
    onColor: (hexColor: string) => {
      // custom color conversion logic here 
      // and return cmyk or rgb
      return cmyk(0,0.78,0.51,1)
    },
    font,
    lang: options.lang,
    title: 'pdfme',
  },
  plugins: getPlugins(),
});
```

### DIFF
--- a/packages/schemas/src/utils.ts
+++ b/packages/schemas/src/utils.ts
@@ -1,5 +1,5 @@
 import type * as CSS from 'csstype';
-import { cmyk, degrees, degreesToRadians, rgb } from '@pdfme/pdf-lib';
+import { cmyk, degrees, degreesToRadians, rgb, Color } from '@pdfme/pdf-lib';
 import { Schema, mm2pt, Mode, isHexValid, ColorType } from '@pdfme/common';
 import { IconNode } from 'lucide';
 import { getDynamicHeightsForTable as _getDynamicHeightsForTable } from './tables/dynamicTemplate.js';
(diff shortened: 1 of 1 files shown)


hand-dot:
@necessarylion 
Thank you for your always interesting PRs. I understand what you are trying to achieve with this one.

Allow me to provide some constructive feedback.

I’m not entirely confident that this approach is the right one, so I’m hesitant to merge it as is. Specifically, I’m concerned that adding the `onColor` option to every schema will increase the complexity of the schema. As you have noted in other issues, SVG and barcodes don’t support CMYK, and we want to keep the schema implementations as simple as possible.  
https://github.com/pdfme/pdfme/issues/460#issuecomment-2262089886

For example, the addition of a [post-processing feature ](https://github.com/pdfme/pdfme/issues/391) might allow us to tackle these issues at a more abstract level. Although I haven’t tested it myself, I’ve seen someone running Ghostscript on Wasm:  
https://meyer-laurent.com/playing-around-webassembly-and-ghostscript

Using Ghostscript for color conversion in post-processing could provide more flexibility. It might even open the door to adding features like PDF/X or PDF/A compliance.

What do you think about this approach?

necessarylion:
@hand-dot Thanks for the review. 

I agree that [pre-processing feature](https://github.com/pdfme/pdfme/issues/391) might solve the issue for SVG and barcodes don’t support CMYK. 

As for using Ghostscript for color conversion, we’ve tried it with a backend service, but the output isn’t always a perfect match, especially when the base template PDF includes transparency objects.

If we extend the schema to support multiple color formats (not just hex codes, but also CMYK and RGB), it could help address the color issue alongside the pre-processing feature. Hex codes could be converted to CMYK or RGB with pre-processing, and during the PDF rendering process, color conversion could be skipped if the color is already specified in CMYK or RGB.

```ts

const template = {
  schemas: [
    {
      headline: {
        ...
        type: "text",
        fontColor: cmyk(54,29,0,35), // or rgb(76,118,166)
      },
    },
  ],
  basePdf: "base64pdf...",
};
```

hand-dot:
@necessarylion 
Ah, I personally think the new proposal is a better implementation. 
It doesn't affect existing users, and it also requires less code modification.

Issue #630, opened 2024-11-13
Designer cannot load PDFs in array/buffer


PlanetTeamSpeakk:
### Describe the bug

When attempting to load a template with a base PDF in the form of a Uint8Array or ArrayBuffer, pdfme seems to wrongly think this is a base64 string and will attempt to parse it as such resulting in an error saying `<some minified name>.split is not a function` (full error below). Only solution is to convert the data into a base64 string.

### To Reproduce

I am using the following function to load the template:
```js
  function loadBasePDF(e: Event) {
    const file = (e.target! as HTMLInputElement).files![0];
    if (!file) return;

    const reader = new FileReader();
    reader.onload = () => {
      const a = new Uint8Array(reader.result as ArrayBuffer);
      // Not including the next line and instead passing the array directly (either as an array or a buffer)
      // will result in the attached error. Converting it to base64 first solves this issue.
      // Obviously, loading the data directly as base64 using readAsDataURL works too.
      const b64 = 'data:application/pdf;base64,' + btoa([...a].map(b => String.fromCharCode(b)).join(''));
      designer.value?.updateTemplate({
        basePdf: b64,
        schemas: [],
      })
    };
    reader.readAsArrayBuffer(file);
  }
  ```

### Expected behavior

PDFMe should be able to parse PDFs in Uint8Array or ArrayBuffer form as well, but it can only parse base64 without errors.

### Your Environment

```markdown
- pdfme package(@pdfme/generator or @pdfme/ui): @pdfme/ui
- pdfme version: 5.2.1
- Operating system: Windows
- Browser name & version: Edge 131.0.2903.36
```

### Your Error Log

```shell
Uncaught (in promise) TypeError: e10.split is not a function
    at kj (index.es.js:9008:15)
    at Lue (index.es.js:26766:13)
    at Nue (index.es.js:26781:33)
    at async index.es.js:130089:15

Which happens on this line:
const t = e.split(";base64,")[1] ? e.split(";base64,")[1] : e, r = Z2e(t), n = new Uint8Array(r.length);
```

hand-dot:
I was handled by @necessarylion .
Released as 5.2.11

PR #659, opened 2024-12-14
[630]: Bug fix on basePdf as array buffer

necessarylion:
[**#630:** Designer cannot load PDFs in array/buffer](https://github.com/pdfme/pdfme/issues/630)


hand-dot:
reviewing

hand-dot:
Hey @necessarylion !!
I discovered an issue where passing an ArrayBuffer to `updateTemplate` would result in it being converted into an empty object by `cloneDeep`, causing it to malfunction.

When I changed `cloneDeep` to `structuredClone`, the tests began to fail, so I will fix that. I will handle the remaining tasks and proceed to the release.

Thank you for your help in fixing the bug!

hand-dot:
Relased as 5.2.11

PR #665, opened 2024-12-30
Allow change the locale of datepicker

kyasu1:
Hi there,

As mentioned on the Discord comment,

https://discord.com/channels/933602537808990268/933602537808990271/1321660507484196944

sometimes we want to display the date picker with a language other than that of UI Designer, which is more developer-oriented. This patch attempts to reflect the `lang` option settings in the schema of date/dateTime components when rendering the AirDatepicker that is used under the hood.

This is still incomplete. The following features need to be addressed:

1. Add a language selection component in the UI panel.
2. The "close" button is still controlled by the language setting of UI Designer. We probably need to somehow import `getDic`t defined in `../ui/src/i18n.ts`.

I am not proficient in TypeScript / Zod development, so please feel free to modify and improve the code as necessary. 

With this patch, loading the next schema, the dateTime picker is rendred with Chinese characters regardeless of the language setting in UI Designer.

```
{
  "schemas": [
    [
      {
        "name": "date-time",
        "format": "yyyy/MM/dd HH:mm",
        "type": "dateTime",
        "content": "2024/12/30 18:33",
        "position": { "x": 130.87, "y": 207.6 },
        "width": 50,
        "height": 10,
        "rotate": 0,
        "alignment": "left",
        "fontSize": 13,
        "characterSpacing": 0,
        "fontColor": "#000000",
        "backgroundColor": "",
        "opacity": 1,
        "required": false,
        "lang": "zh"
      }
    ]
  ],
  "basePdf": {
    "width": 210,
    "height": 297,
    "padding": [20, 20, 20, 20],
    "staticSchema": []
  },
  "pdfmeVersion": "5.2.11"
}
```


hand-dot:
Hey @kyasu1 

Thank you for your work—I see what you’re trying to achieve. I have two quick comments:

1. You’ve added some unsupported languages to the `locales` array in pdfme. Is this strictly necessary? I believe the languages listed here should suffice:  
   https://github.com/pdfme/pdfme/blob/3e7854eca3353960c8d305211f636809268fb00a/packages/common/src/schema.ts#L3

2. The date format options are also tied to the designer’s language settings. It would be ideal to use the locale from this PR to dynamically update those options as well.
  CleanShot 2025-01-02 at 18 10 47@2x


Would you be able to make these adjustments on your end? If any of my feedback is unclear, please let me know. 
I’m happy to support you and help bring this to release as quickly as possible.

If the changes are too challenging to implement right now, feel free to let me know and I can take over. 

In any case, thank you for kicking off the new year with a great PR!

kyasu1:
Thank you for your reply.

Regarding the first point, I believe the languages defined in pdfme as `Lang` are intended for displaying field labels in the UI Designer, which is beneficial primarily for developers and designers. However, these should be considered distinct from the languages used in components defined as `Locale`. By supporting all locales defined in AirDatepicker, the next scenario will be possible:

> A programmer in Thailand who is working for a Canadian French company, is able to use the UI Designer in Thai to program / design but display the calendar in Canadian French style.

Regarding the second point, defining possible datetime formats for all locales is quite challenging. The current design presents some common formats as options in the select component. I suggest abandoning this approach and instead making this field editable, allowing developers to write a datetime format that follows the rules defined [here](https://air-datepicker.com/docs?scrollTo=dateFormat). I am not familiar with `XRender`, but if it supports a combobox component we may be able to support both functionalities.

I have managed to implement this approach, and I will update the code later after sorting things out.

kyasu1:
I have pushed modifications to make the date format field editable. Not sure this is the best approach but looks working well.

hand-dot:
@kyasu1 Cool, I will check when i have time.
Please wait a while.

packages/schemas/src/date/helper.ts
```diff
@@ -139,17 +129,28 @@ export const getPlugin = ({
         }
       };
 
+      const formatted = (date: string | Date): string => {
```
hand-dot, line 132: this one is never used.

packages/schemas/src/date/locale.ts
```diff
@@ -0,0 +1,95 @@
...
+import localeZh from 'air-datepicker/locale/zh';
+
+import * as dateFns from 'date-fns/locale';
+
+export const locales = [
+    'ar', 'bg', 'ca', 'cs', 'da', 'de', 'el', 'en', 'es', 'eu', 'fi', 'fr', 'hr', 'hu', 'id', 'it', 'ja',
+    'ko', 'nb', 'nl', 'pl', 'pt-Br', 'pt', 'ro', 'ru', 'si', 'sk', 'sl', 'sv', 'th', 'tr', 'uk', 'zh',
+] as const;
+const localeSchema = z.enum(locales);
+export type LocaleEnum = (typeof locales)[number];
```
hand-dot, line 46: no one use this. pls remove

packages/schemas/src/date/helper.ts
```diff
@@ -273,6 +280,15 @@ export const getPlugin = ({
               },
             ],
           },
+          locale: {
```
hand-dot, line 283: If my understanding is correct, this is a Locale used only for the UI, right?  
If so, it might be better to rename it to something like `localForUI`.

packages/common/src/schema.ts
```diff
@@ -79,6 +80,7 @@ export const Dict = z.object({
   'schemas.table.columnStyle': z.string(),
 
   'schemas.date.format': z.string(),
+  'schemas.date.locale': z.string(),
```
hand-dot, line 83: Perhaps `localForUI` might be more appropriate.

hand-dot (commented):
@kyasu1  Thank you.  
I believe this change increases formatting flexibility and benefits many users.  
I’ve added a few comments. Please feel free to ask if there’s anything unclear or you’d like to discuss further.

packages/schemas/src/date/helper.ts
```diff
@@ -273,6 +280,15 @@ export const getPlugin = ({
...
+            title: i18n('schemas.date.locale'),
+            type: 'string',
+            widget: 'select',
+            props: {
+              options: Locale.locales.map((locale) => ({ label: locale, value: locale })),
```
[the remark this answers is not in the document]
kyasu1: About the first point, how about adding national flags and locale names in that language like below? I think this is easy to implement. Also, changing the language used in the locale names depending on the lang of the Form/Viewer might be possible, but I think that is a bit overkill unless users really need them.

> 🇺🇲 English
> 🇯🇵 日本語
> 🇨🇳 繁体字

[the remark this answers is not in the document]
kyasu1: The second point is related to [this](https://github.com/pdfme/pdfme/pull/665#discussion_r1906236437) comment, i.e., how the locale field of a newly placed component should be handled.

I think the best scenario would be something like this:

1. Select a `lang` from the Designer menu.
2. Place a new DateTime component.
3. The locale field should be set to the same value as `lang` if it exists.

This way, designers won't need to pick the locale field each time they place a new component.

If the `lang` value is not supported in the `locale` menu, its value will be set to `en` as the default.

However, to achieve this behavior, I think we would need to pass the `options.lang` value to `defaultSchema`, which might affect other plugins. Therefore, I'm not sure if this is the best way to proceed.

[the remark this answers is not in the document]
kyasu1: > About the first point, how about adding national flags and locale names in that language like below? I think this is easy to implement. Also, changing the language used in the locale names depending on the lang of the Form/Viewer might be possible, but I think that is a bit overkill unless users really need them.
> 
> > 🇺🇲 English
> > 🇯🇵 日本語
> > 🇨🇳 繁体字

Umm flag icons might not be a good idea, i18n is hard :s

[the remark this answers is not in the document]
hand-dot: Hey @kyasu1  Thank you.

- I personally think the flag emojis are a bit too much, but overall, I like the idea.  
- If the schema doesn't explicitly specify a locale, I think simply falling back to `option.lang` should suffice.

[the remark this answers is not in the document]
kyasu1: I have thought carefully about this section again. 

The people who will primarily use this list item are developers. Therefore, I believe the current setting is easily recognizable because it is expressed using a two-letter alphabetical representation that conforms to ISO 639-1, which is familiar to them.

The representations, such as the ones I suggested previously, are possible. However, I feel they are redundant for developers.

- ar (عربي)
- bg (български)
- ca (català)
- en (English)
- ja (日本語)

packages/schemas/src/date/helper.ts
```diff
@@ -194,39 +186,57 @@ export const getPlugin = ({
...
+              },
+            ],
             span: 24,
           },
           fontName: {
             title: i18n('schemas.text.fontName'),
             type: 'string',
             widget: 'select',
             default: fallbackFontName,
+            placeholder: fallbackFontName,
```
hand-dot, line 239: ❤️

hand-dot:
@kyasu1 Thank you. I’m running the code locally and reviewing it again. 

There’s a possibility I might make a few adjustments on my end, but I truly appreciate the excellent work you’ve done.

I believe I’ll be able to merge it into the main branch within a few hours.

hand-dot:
I found some issues while working on it, so it seems it will take more time before merging.

packages/schemas/src/date/helper.ts
```diff
@@ -56,100 +147,124 @@ const injectStyles = (css: string) => {
...
         height: `${schema.height}mm`,
         display: 'flex',
         flexDirection: 'column',
         justifyContent: mapVerticalAlignToFlex(VERTICAL_ALIGN_MIDDLE),
-        position: 'relative',
       };
+      Object.assign(textElement.style, textElementStyle);
 
-      const input = document.createElement('input');
+      await text.ui(
```
hand-dot, line 230: By rendering using `text.ui`, you can intentionally display Tofu in the browser even when the font is unavailable.

packages/schemas/src/date/types.ts
```diff
@@ -9,4 +9,5 @@ export interface DateSchema extends Schema {
   characterSpacing: number;
   fontColor: string;
   backgroundColor: string;
+  locale?: string;
```
hand-dot, line 12: Sorry, I ended up reverting `localeForUI` back to `locale`.

I realized that when the format is something like `MM/dd/yyyy hh:mm aa`, `localeForUI` is no longer just a property for UI since it is also used to determine the `aa` part of the value.

packages/schemas/src/date/helper.ts
```diff
@@ -1,28 +1,48 @@
 import type * as CSS from 'csstype';
-import AirDatepicker, { AirDatepickerLocale } from 'air-datepicker';
+
+import AirDatepicker from 'air-datepicker';
+import type { AirDatepickerLocale } from 'air-datepicker';
+import localeAr from 'air-datepicker/locale/ar';
```
hand-dot, line 5: I think it's a very ugly import. However, I combined everything into a single file because it makes refactoring easier with tools like Cline or ChatGPT.

packages/schemas/src/date/helper.ts
```diff
@@ -165,68 +279,94 @@ export const getPlugin = ({
...
-
         const fontNames = Object.keys(font);
         const fallbackFontName = getFallbackFontName(font);
 
+        const locale = getAirDatepickerLocale(
+          (activeSchema as any).locale || options.lang || defaultLocale
+        );
+
+        if (
+          (activeSchema as any).locale === undefined &&
```
hand-dot, line 320: Since the initial value of `locale` is `undefined`, the following `if` block is executed during the first render, allowing the `locale` and `format` to be dynamically set using the current `options.lang`.

If `locale` is explicitly specified, this `if` block will not be executed, allowing users to freely configure the `locale`.

hand-dot:
@kyasu1 !
Very nice work!!

hand-dot:
Published pdfme v5.2.12!

Also we can play from here: https://playground.pdfme.com/
CleanShot 2025-01-18 at 16 31 02@2x

packages/schemas/src/date/helper.ts
kyasu1: @hand-dot Nice work, this is what I wanted to achieve the first time! The usability has become much more intuitive.

hand-dot:
@kyasu1 
I made some adjustments to the behavior and labels of the datepicker by this [PR](https://github.com/pdfme/pdfme/pull/679), so I wanted to report them just in case.  

Specifically, I decided to use `options.lang` to handle i18n for the "clear" and "set" datepicker buttons.  
I felt it was more natural to use `locale` only to control the calendar and format.

PR #675, opened 2025-01-14
Add textIndent property to text schema and update related components

anAverageSlavGuy:
I have urgent need of this feature cause my text boxes have indent on the first row.
This is a common feature of all text paragraphs, like in magazines or in books.

image


peteward:
Hi there!

Thanks for the contribution. 

It looks like you've implemented the UI side in em, but not the pdf rendering. You'll need both and will want to use an approach that allows the spacing used to be aligned between them (check out the way other positioning is done).

Having said that, I'm not sure if this would be considered a core text attribute. The text plugin will start to get really bloated for the user if there are lots of extra attributes added for various use cases. Perhaps a better option would be to write your own plugin which extends the text one and adds your changes? There are a couple of examples of schemas that extend or use one another's functions that you could try to copy.

anAverageSlavGuy:
Hi, ok thanks, it might be a good idea to create my own custom plugin, I don't quite understand how I can then import it from @pdfme/schemas like the other existing plugins... it means that I still have to add it to the schemas package and add it to the repo designer or do I have to use my own personal playground?

peteward:
hmm actually, it might not be as easy as I thought to extend the text plugin.

If you want to create your own plugin that is standalone then it's easy. You don't need to add it to the schemas package, it can existing your own application or a standalone package and just pass it to pdfme. 

However, if you want to modify an existing plugin then there's no easy way to hook into them to make adjustments. You would either have to copy&paste all of the text plugin into your own and make your changes, or fork the project.

anAverageSlavGuy:
Ok so it's as I had thought, unfortunately I have to make a separate plugin, I just needed this little extra functionality which in my opinion could also be a core feature of the text, given that multiline texts are often used. No problem, I'll proceed another way, thanks!

peteward:
Apologies, it's up to @hand-dot whether he thinks this should be a core text feature.

hand-dot:
I understand what you're trying to achieve, but at the moment, I don't think this should become a core feature of the text module. Addressing this issue within the library is complex, but the problem can be resolved with minimal sacrifices in design.

Therefore, I recommend addressing it by creating a custom plugin. In this case, it would be best to copy the existing text schema (located at `packages/schemas/src/text`) into your application and modify the code as needed.

Additionally, regarding the modifications to the PDF rendering process, properties like `text-indent` are not available, so you will need to calculate them manually.

hand-dot:
I'm glad you created this PR. Thank you, @anAverageSlavGuy
But I will close this PR for now due to above reason.

If you have any updates or questions, please let me know on Discord. I'll do my best to respond.

PR #677, opened 2025-01-16
Fixed index calculation on schema render

jgrana:
This fixes tab index on Preview


hand-dot:
Hey @jgrana !
Thank you for creating PR.

I cloned your code and ran it locally, but I couldn't figure out what issue it is trying to solve. 
From what I observed, the original code seems to handle `tabIndex` in the Form correctly.

Could you clarify the specific case where the issue occurs and how your changes improve the situation? 
It would be helpful if you could include images or other visuals to explain it. Thank you!

hand-dot:
I found some issues while working on it, so it seems it will take more time before merging.

hand-dot:
no response

jgrana:
if there are more than 2 pdf pages the tabIndex is broken

PR #679, opened 2025-01-18
Fix DatePicker

hand-dot:
- Address the issue where the datepicker closes when selecting a time in `time` or `dateTime` mode
- Modify the datepicker’s button labels to use `i18n.ts` (not calender)
  - It felt unnatural to handle multilingual support within the schemas

### DIFF
--- a/packages/common/src/schema.ts
+++ b/packages/common/src/schema.ts
@@ -7,6 +7,8 @@ export const Dict = z.object({
   // -----------------used in ui-----------------
   cancel: z.string(),
   close: z.string(),
+  set: z.string(),
+  clear: z.string(),
   field: z.string(),
   fieldName: z.string(),
   align: z.string(),
--- a/packages/schemas/src/date/helper.ts
+++ b/packages/schemas/src/date/helper.ts
@@ -60,62 +60,43 @@ type PickerType = 'date' | 'time' | 'dateTime';
 interface Locale {
   label: string;
   adLocale: AirDatepickerLocale;
-  close: string;
   formatLocale: dateFns.Locale;
 }
 
-const LOCALE_MAP: Record<
-  string,
-  {
-    label: string;
-    adLocale: AirDatepickerLocale;
-    close: string;
-    formatLocale: dateFns.Locale;
-  }
-> = {
-  ar: { label: 'Arabic', adLocale: localeAr, close: 'يغلق', formatLocale: dateFns.ar },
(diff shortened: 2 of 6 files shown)


kyasu1:
Hmm, I think this change broke the buttons' label translation again... They are now following the Lang value, not the locale value, so the result is very weird as shown.

Screenshot 2025-01-18 at 21 31 19

Screenshot 2025-01-18 at 21 34 54

kyasu1:
Moreover, the datePicker is no longer closing when clicking outside, but we always need to click the Set button. This behavior is very frustrating. This happens in both the Designer and Form modes.

Screenshot 2025-01-18 at 21 39 35

kyasu1:
Also, since you are specifying 'ja' as the `lang` value in the following line,

https://github.com/pdfme/pdfme/blob/2ff80d1d557eabe618708b27440b4ff94d17653c/playground/src/FormAndViewer.tsx#L61

the users will always see the buttons' labels in Japanese in Form mode, regardless of the `Lang` or `Locale` values.

I understand that the playground is just an example of how to use this library. However, this inconsistency is very confusing for new users.

hand-dot:
@kyasu1 Thank you for checking again.

Yes, maybe this playground form move should be able to change lang option.

I guess this makes misleeding.

hand-dot:
at least we need to sepalete issue playdground and date kind schema.

I will update playground form mode lang option selectable.

hand-dot:
Update Playground: https://playground.pdfme.com/form-viewer
CleanShot 2025-01-19 at 12 53 07@2x

kyasu1:
Thank you for the quick update! It looks like it's working well. However, I think the buttons' labels should reflect the `Locale` value rather than the `Lang` value. As shown in the next screenshot, the calendar is in Russian, but the buttons are in Thai. This looks very weird.

Screenshot 2025-01-19 at 13 02 16

hand-dot:
Hmm, I see. Shall we create a translation adapted for the calendar?  
If you handle the task, I will review it right away.

kyasu1:
The following lines are what I suggested first time:

https://github.com/pdfme/pdfme/blob/37a6e1f70eb3e75105f8b30ddfd3536db160348e/packages/schemas/src/date/helper.ts#L76-L96

I think hard coding the translations of buttons here is valid, since they are usable in this schema only which depends on the external module. In this way, change of `Lang` on UI (Designer  / Form) will not affect to the datePicker language. The `Lang` value should be used only when placing a new component as you implmented. So I suggest not to use `i18n` in this case and revert to `5.2.13`...

Issue #638, opened 2024-11-18
@pdfme/generator generate() takes a blank space as a line break when CJK


induang:
### Describe the bug

I designed a pdf editor that users input some japanese chars and then show in current tab(new Viewer) or preview it in new tab(generate()).
and it seems that generate()  takes a blank as a line break, which do not happen when i use new View. it can just simply reproduce in official web.
Although I fixed it by changing all blank by \u00A0, but it always be more great then i do not need to handle. 

### To Reproduce

<img width="1886" alt="Screenshot 2024-11-18 at 16 26 55" src="https://github.com/user-attachments/assets/0cea2cc2-84ac-4ba2-9c0c-288b7193ded6">
<img width="1537" alt="Screenshot 2024-11-18 at 16 28 08" src="https://github.com/user-attachments/assets/5e708014-ab8b-4590-bec7-0805338a0239">

### Expected behavior

behavior as same as new Viewer()

### Your Environment

```markdown
offcial web.
chrome lastest
mac air 2
```

### Your Error Log

```shell
nothing.
```

peteward:
Hi @induang,

Please can you share the exact input string that is being passed to generate?

Thanks

induang:
Sure! @peteward 
アプリをお使いにならない方は、解錠パスワードを使用して解錠することができます。解錠パスワードは1法人IDに対して1つです。

peteward:
Thanks.

So it seems that browsers implement a complex set of language-specific rules around line breaking. There are specialist engines that do this. 

In pdfme the implementation is a very simple version designed for Western character sets (breaking on all white space).

Whilst it might be possible to hardcode some additional rules checking for CJK characters in our algorithm, it would probably be better to try a more comprehensive library or engine for determining the line-breaking. 

I don't think this is a quick fix...

kyasu1:
Hi @induang, @peteward and @hand-dot !

[`Intl.Segmenter`](https://developer.mozilla.org/ja/docs/Web/JavaScript/Reference/Global_Objects/Intl/Segmenter)  is now available in all major browsers. To solve this issue, I think we can use it for the line-breaking alghorithm instead of the following function.

https://github.com/pdfme/pdfme/blob/c8f2057af816864673c7c0c956acecf2d1e1f47e/packages/schemas/src/text/helper.ts#L199

I wrote a rough implementation and got the following result.

![Image](https://github.com/user-attachments/assets/d437d767-35a6-42f7-ac0f-e199bd48bf36)

[...]

PR #686, opened 2025-01-26
use Intl.Segmenter for line breaking and support  alignment

kyasu1:
This will solve issue [#638](https://github.com/pdfme/pdfme/issues/638) and include the following enhancements:

1. Use `Intl.Segmenter` for breaking text into lines in all languages.
2. Add `locale` property to the text schema
3. Add `locale` property to the table schema but not much tested
4. Add `justify` alignment to the text schema. This is very common in CJK languages.
5. Add simple `禁則処理` or Japanese hyphenation, which prevents unappealing formatting.

Some tests are also added as `integration5.test.ts` under the `generator` directory.

Regarding performance, there is no significant degradation in testing.

### Original

```
> @pdfme/generator@0.0.0 test
> jest

 PASS  __tests__/integration3.test.ts
 PASS  __tests__/integration4.test.ts (6.928 s)
 PASS  __tests__/integration2.test.ts (7.742 s)
 PASS  __tests__/integration1.test.ts (7.835 s)
 PASS  __tests__/generate.test.ts (8.653 s)

Test Suites: 5 passed, 5 total
Tests:       58 passed, 58 total
Snapshots:   0 total
Time:        8.851 s, estimated 9 s
Ran all test suites.
```

### Intl.Segmenter

```
> @pdfme/generator@0.0.0 test
> jest

 PASS  __tests__/integration5.test.ts
 PASS  __tests__/integration3.test.ts
 PASS  __tests__/integration4.test.ts (5.118 s)
 PASS  __tests__/integration1.test.ts (7.809 s)
 PASS  __tests__/integration2.test.ts (7.894 s)
 PASS  __tests__/generate.test.ts (8.665 s)

Test Suites: 6 passed, 6 total
Tests:       60 passed, 60 total
Snapshots:   0 total
Time:        8.883 s, estimated 11 s
Ran all test suites.
```

To pass the tests, I needed to recreate most of PDFs under the tests directory, that bloats the commit...


kyasu1:
Screenshot 2025-01-26 at 22 54 33

peteward:
Thanks for the contribution!

I see that you have named some of these variables `locale`, but shouldn't it be `language` as I see you are using 2 character strings such as `en` and `ja`?

packages/schemas/src/text/propPanel.ts
```diff
@@ -151,6 +153,17 @@ export const propPanel: PropPanel<TextSchema> = {
           },
         ],
       },
+      locale: {
```
peteward, line 156: Adding `locale` to every the prop panel feels quite heavy, and often it will be redundant - I doubt that most users will know what it does.

Should this not be a top-level option for the Designer that you can pass in, and/or captured from the users browser?

It's possible that you might want to use different locales in a single PDF but feels very unlikely to me... Perhaps there could be an option to enable this level of granular control though?

So we could do:

1. Detect locale from user browser when using the Designer
2. Override with an option to force the locale when creating the Designer component
3. Override with an option that allows granular control of the locale on every text component

wdyt?

We might also need to think about migrating legacy schemas if a different locale from `en` is being used (we have a mechanism for doing schema migrations already).

packages/schemas/src/text/pdfRender.ts
```diff
@@ -151,9 +150,11 @@ export const pdfRender = async (arg: PDFRenderProps<TextSchema>) => {
   }
 
   const pivotPoint = { x: x + width / 2, y: pageHeight - mm2pt(schema.position.y) - height / 2 };
+  const segmenter = new Intl.Segmenter(schema.locale, { granularity: 'grapheme' });
```
hand-dot, line 153: Define this variable close to where it will be used.

This will be around bellow if block

> if (alignment === 'justify' && (lines[rowIndex + 1] ?? '' !== '')) {

packages/schemas/src/text/propPanel.ts
hand-dot: I feel the same way. 

Switching the `locale` per schema is a rare use case, and I’d like to keep the schema settings as simple as possible. The UI package is not designed just for programmers. The UX should be intuitive enough that even someone’s non-tech-savvy mother could create a template effortlessly.

Therefore, how about adding `intlSegmenterLocale` to `CommonOptions` like this?  
(It might also be a good idea to create a type for a string that adheres to BCP 47 language tags.)

```ts
export const CommonOptions = z.object({ 
  font: Font.optional(),
  intlSegmenterLocale: z.string().optional(),
 }).passthrough();
```

https://github.com/kyasu1/pdfme/blob/b1d6107e2004058ce4bc2766927c52373ef65295/packages/common/src/schema.ts#L157

For the `date` and `table` schemas, they should continue to call the `text` schema without being aware of the `locale`. Within the `text` schema, we can access the optional `intlSegmenterLocale` option and determine the locale for `Intl.Segmenter`.

hand-dot:
Hi @kyasu1 
Thank you for creating the PR.

I found a discrepancy between the UI and the PDF generated by the generator.

CleanShot 2025-01-28 at 10 47 16@2x

CleanShot 2025-01-28 at 10 47 40@2x

Even if it’s difficult to achieve a perfect match, this could confuse users, so it needs to be fixed.

---

I’ll provide as much support as possible until the merge, so feel free to ask if you have any questions.

packages/schemas/src/text/pdfRender.ts
kyasu1: With that,`segmenter` will be created multiple times in the loop. Isn't that inefficient? Or no penallty on JS engines?

hand-dot: Ah, got it. pls ignore this comment.

kyasu1:
> Hi @kyasu1 Thank you for creating the PR.
> 
> I found a discrepancy between the UI and the PDF generated by the generator.
> 
> Even if it’s difficult to achieve a perfect match, this could confuse users, so it needs to be fixed.
> 

On Adobe illustrator, there are multiple possibilities for `justify` alignment like below:

Screenshot 2025-01-28 at 12 16 24

Screenshot 2025-01-28 at 12 16 31

The current behavior is closer to the second option, but its representation in CSS might not be straightforward.

In general, the first option is more appropriate for most languages, so I think we should implement that behavior. Probably we can achieve this by not justifying if the text is a single line. Let’s give it a shot.

packages/schemas/src/text/propPanel.ts
kyasu1: Thank you for your replies.

I've been experimenting with this text schema by changing the `locale` value to various strings. However, I realized that there is no visible difference regardless of the `locale` string specified. For example, if you write a Japanese text and then change the `locale` to `zh`, `en`, or any other value, the generated PDFs remain the same and looks correct.

After doing some research, I came across this Stack Overflow question.

https://stackoverflow.com/questions/75747868/how-exactly-locale-param-affects-the-result-of-intl-segmenter-execution-in-jav

It seems that `Intl.Segmenter` determines how to segment text based on the code points of the text itself. In very rare cases, the `locale` value may affect the result. Therefore, providing either `undefined` or `en` as the `locale` to `Intl.Segmenter` seems to work fine in most cases.

So, should we consider dropping the locale from the text schema? What about hyphenation? I've implemented simple Japanese-specific rules, which are also based on the code points of Japanese characters. If we were to implement similar rules for other languages, they would likely also rely on the code points of those languages. However, if the locale is not included in the schema, we would need to test all language-specific hyphenation rules, which would be inefficient. Therefore, retaining the locale in the schema still seems appropriate.

That said, I am still wondering what the best way to handle this locale issue might be...

hand-dot: Do you know the following library?  
[https://www.npmjs.com/package/linebreak](https://www.npmjs.com/package/linebreak)  

I haven't used it before, but it might work well in this case.  
Since it processes line-breaking rules based on Unicode, it should support not only Japanese but also Chinese and other languages.

What do you think

hand-dot: In other words, I believe that by delegating the current line-breaking logic, which handles spaces and hyphens, to the `linebreak` library, it could support a wider range of languages.  
[https://github.com/pdfme/pdfme/blob/0a4d425b0e24254617b31e6d52b4703f6fb1ac68/packages/schemas/src/text/helper.ts#L166](https://github.com/pdfme/pdfme/blob/0a4d425b0e24254617b31e6d52b4703f6fb1ac68/packages/schemas/src/text/helper.ts#L166)  

The PR should become much simpler.

kyasu1: @hand-dot 

We should stay with `Intl.Segmenter`, which is very clever and works well without specifying the locale explicitly. So, I suggest that we remove `locale` from the schema until it becomes really necessary. One drawback is the `禁則処理` loop need to be run every time unconditionaly. 

Making the line breaking algorithm switchable is good idea though. We might want to use LLM tokenizer.

kyasu1: I have just now tested the `linebreak` library. However, it merely splits Japanese sentences on a character-by-character basis, which is not what we are looking for.

hand-dot: > Making the line breaking algorithm switchable is good idea though. We might want to use a LLM based tokenizer.
Interesting. I think https://github.com/google/budoux is good choice.
But, I love to use simple and light approach

> I have just now tested the linebreak library. However, it merely splits Japanese sentences on a character-by-character basis, which is not what we are looking for.
My understanding is that this library can be used to find possible line break positions.

CleanShot 2025-01-29 at 12 55 22

kyasu1:
@hand-dot  could you take a look at the updated commits? 

- Fixed the bug related to `justify` in short English texts. This disables `均等揃え` or `lastJustify`, but we can implement it later if necessary.  
- Removed `locale`. `Intl.Segmenter` seems to be less sensitive to this value, and the generated PDFs showed no visual differences regardless of the specified value. Unless someone specifically requires this property, we can omit it for now.  
- Some snapshot tests are failing because the PDFs have not been updated. This was done intentionally to keep the changes in this commit minimal.

hand-dot:
Hi @kyasu1 
Thank you.  
I am currently checking. I will report back if there are any additional comments.  

Since `locale` is no longer set from the schema, could you revert some of the unnecessary changes?  
Please minimize the number of modified files as much as possible.

packages/schemas/src/text/constants.ts
```diff
@@ -21,3 +22,83 @@ export const DEFAULT_DYNAMIC_MIN_FONT_SIZE = 4;
 
 export const DEFAULT_DYNAMIC_MAX_FONT_SIZE = 72;
 export const FONT_SIZE_ADJUSTMENT = 0.25;
+
+export const LINE_START_FORBIDDEN_CHARS = [
```
hand-dot, line 26: I'm looking for a better way to manage such rules.  
This issue occurs not only in Japanese but also in various languages such as Chinese, Korean, and Thai.  
If there is a dataset that manages these characters, I would like to use it.

packages/generator/__tests__/assets/pdfs/assert/segmenterEnglish.pdf
kyasu1, line 1: Adobe Acrobat on Mac wont open this file. No probelm on browsers and preview on Mac OS.

kyasu1:
Hi @hand-dot 

Removed unused locale related codes as told. Also updated assert PDFs.

Added a fix for handling of whitespace or line-breakable characters at the beginning of a line when wrapped.

On the handling of very long words, [dynamicFontSizeVertical.json](https://github.com/pdfme/pdfme/blob/main/packages/generator/__tests__/assets/templates/dynamicFontSizeVertical.json), there is a discripancy between UI (left) and PDF (right) as follows.

Screenshot 2025-01-30 at 18 34 03

And one more isssue, Adobe Acrobat raises an error when opening the following generated PDF file:

```
packages/generator/__tests__/assets/pdfs/assert/segmenterEnglish.pdf
```

packages/schemas/src/text/constants.ts
kyasu1: Probably we need a way to handle language-specific rules, but that might reintroduce `locale`...

Some characters are common, but others are not.

The following blog posts by Morisawa are very informative. We may be able to define four rules as in Adobe InDesign, then create a set of rules for each language.

https://note.morisawa.co.jp/n/n64589ad7572d

However, this can be quite challenging without support from native speakers.

hand-dot:
Hi @kyasu1 

I have confirmed the fix and can see the improvements.  

However, I found an issue where some text overflows from the box when the alignment is set to **left**. This does not occur when the alignment is **justify**.  

This issue happens in both Japanese and English.  
Can you fix this problem?


CleanShot 2025-02-04 at 10 44 33@2x


CleanShot 2025-02-04 at 10 44 44@2x

Reproducible template: [template (8).json](https://github.com/user-attachments/files/18650483/template.8.json)

kyasu1:
@hand-dot thank you for checking and pointing out. This should fix the issue. Please take a look at again.

hand-dot:
@kyasu1 
alignment": "right of `multiVariableText.pdf` looks wrong.


CleanShot 2025-02-05 at 09 33 39@2x

CleanShot 2025-02-05 at 09 33 48@2x

kyasu1:
Hi @hand-dot , this will fix the issue you mentioned. 

The character spacing was not took into account in the calculation which caused that result. 

By updating two assert PDFs, all tests passed. I have checked all generated PDFs in Adobe Acrobat Reader, and they look okay except for pdfImage.json, in which no image was shown. It seems unrelated to this update since it would also not be shown on pdfme.com`

As you may notice, there is discrepancy between HTML/CSS and the generated PDF, as shown in the attached image. The text is wrapped at the wrong place in the PDF, with the same result on pdfme.com` Also you may also notice slight differences in another templates.

Screenshot 2025-02-06 at 19 31 03

kyasu1:
I think we need to discuss one more thing, that is how to handle trailing spaces. The current `pdfme.com` removes all trailing spaces when generating PDFs. However, HTML/CSS preserves them as you can see in the Viewer. In the case of right alignment, the spaces somewhat overflow the border but they still exist, except for the last line. As far as I have tested, Adobe Illustrator also follows similar rules. Should we try to implement the same rules?

hand-dot:
Hi @kyasu1 , Thank you for spending your valuable time solving this problem.

> As you may notice, there is discrepancy between HTML/CSS and the generated PDF, as shown in the attached image. The text is wrapped at the wrong place in the PDF, with the same result on pdfme.com` Also you may also notice slight differences in another templates.

maybe we need to check bellow.
https://github.com/kyasu1/pdfme/blob/43c67db2f872bfdc7ca5f2a232945966051275c0/packages/schemas/src/text/uiRender.ts#L218

However, this issue is not caused by this PR but is an existing problem, so it's okay to skip it.  
If you choose to skip it, I will create a separate issue.



> I think we need to discuss one more thing, that is how to handle trailing spaces. The current pdfme.com removes all trailing spaces when generating PDFs. However, HTML/CSS preserves them as you can see in the Viewer. In the case of right alignment, the spaces somewhat overflow the border but they still exist, except for the last line. As far as I have tested, Adobe Illustrator also follows similar rules. Should we try to implement the same rules?

I see. I think it would be better to make it behave the same way as Adobe Illustrator.

kyasu1:
Hi @hand-dot , I just opened a new issue #697 regarding the first problem :-)

hand-dot:
Hi @kyasu1 !

There are some issues pointed out below, but I think it's okay to merge this PR.  
https://github.com/pdfme/pdfme/pull/686#issuecomment-2642822116

Is there anything left to be done? If not, I’d like to proceed with the merge.

kyasu1:
Hi @hand-dot , I found some issues in edge cases and am working on them. Please wait for the next update...

kyasu1:
I was mistaken about the second point regarding trailing spaces. In PDF, trailing spaces should always be trimmed.

Therefore, if text is set to align right, we should ignore trailing spaces and align the entire text to the right boundary. I have tested this behavior in both Adobe Illustrator and Figma, and they follow this convention.

Also, I have fixed some issues that occurred in edge cases. Additionally, I rewrote functions related to Japanese text rules, which were previously written using recursion and contained bugs. The new versions are more verbose but easier to follow and understand.

Some templates have inconsistencies between the HTML/CSS and PDF versions, but I believe they are acceptable at this time.

hand-dot:
@kyasu1 Could you push your new changes?

hand-dot:
@kyasu1 

> Also, I have fixed some issues that occurred in edge cases. Additionally, I rewrote functions related to Japanese text rules, which were previously written using recursion and contained bugs. The new versions are more verbose but easier to follow and understand.

Where is this change?

kyasu1:
@hand-dot, https://github.com/pdfme/pdfme/pull/686/commits/ec7e911c0f26c335405d5e5e02fda2c220903f60

packages/schemas/src/text/helper.ts
```diff
@@ -355,8 +363,172 @@ export const splitTextToSize = (arg: {
...
+};
+
+//
+// 日本語禁則処理
+//
+// https://www.morisawa.co.jp/blogs/MVP/8760
+//
+// 行頭禁則
+export const filterStartJP = (lines: string[]): string[] => {
+  let filtered: string[] = [];
```
hand-dot, line 454: Pls use const

packages/schemas/src/text/helper.ts
```diff
@@ -355,8 +363,172 @@ export const splitTextToSize = (arg: {
...
+          }
+        } else {
+          lines[++lineCounter] = char;
+          currentTextSize = size + characterSpacing;
+        }
+      }
+    }
+  }
+
+  if (true) {
```
hand-dot, line 429: Pls use containsJapanese function

```js
function containsJapanese(text: string): boolean {
  return /[\p{Script=Hiragana}\p{Script=Katakana}\p{Script=Han}]/u.test(text);
}
```

hand-dot:
@kyasu1 Could you check my comment?

hand-dot (approved):
LGTM

hand-dot:
@kyasu1 
Thank you!
Thanks to you, the typography has improved significantly👍

hand-dot:
Released as [5.3.5](https://github.com/pdfme/pdfme/tree/refs/tags/5.3.5)

@kyasu1 Could you check Discord DM?

PR #690, opened 2025-01-30
check file selected or not

kyasu1:
If users click the "cancel" button on the file selector for the first time, the `change` event is not triggered. However, if the `input` element has a previous selection, the event is triggered but the `e.target.files` array is empty, which cause an error in  the subsequent `readFile`.


necessarylion (approved):

hand-dot (approved):
Nice!

PR #691, opened 2025-01-31
`plugin` should be checked before its usage

kyasu1:
With the playground, it fails to open the the following template, since the `ean8` barcode plugin is not installed. That's okay but we should inform users what is wrong instead of cryptic error messages.

https://github.com/pdfme/pdfme/blob/main/packages/generator/__tests__/assets/templates/Aone72230JAN%E3%82%B3%E3%83%BC%E3%83%89%E7%9F%AD%E7%B8%AE.json


hand-dot (approved):
LGTM

Issue #636, opened 2024-11-17
Enable expressions for specific columns in the table


hand-dot:
### Is your feature request related to a problem? Please describe.

Currently, [expressions](https://pdfme.com/docs/expression) cannot be used in table cells, so this functionality should be implemented.

![CleanShot 2024-11-17 at 17 04 19@2x](https://github.com/user-attachments/assets/cc82feae-306b-43dd-8e46-c82e48aaec29)

### Describe the solution you'd like

Allow specific columns to be set as `readOnly` while enabling expressions to be written within those columns.

### Describe alternatives you've considered

none

roopantj:
@hand-dot 
I can utilize expressions within table cells. Has the aforementioned enhancement been implemented?  
If not, could you please provide a detailed explanation of the requirement?

![Image](https://github.com/user-attachments/assets/08c0245c-269c-40f7-af16-b78d1c190993)

![Image](https://github.com/user-attachments/assets/4a153269-b6fc-4bf6-b183-a13797f2d002)

hand-dot:
Hi @roopantj !

1. As you mentioned, if the table is set to read-only, we can use an expression right now.  
2. However, what I want to achieve is to make only specific columns read-only and be able to use expressions, as shown in the image below.  
[...]

PR #694, opened 2025-02-02
Enabled expressions in table cells

roopantj:
Issue [636](https://github.com/pdfme/pdfme/issues/636)

I have enabled expressions in table cells as per the requirement.  
The context is now created based on the row scope for easier accessibility. Let me know if there are any improvements needed.  

PFA screenshots.  

To access data for a specific row, column names can be used within expressions.
 
 Edit Mode
image
View Mode
image

@hand-dot

### DIFF
--- a/packages/schemas/src/tables/pdfRender.ts
+++ b/packages/schemas/src/tables/pdfRender.ts
@@ -1,5 +1,6 @@
 import type { TableSchema } from './types';
 import type { PDFRenderProps } from '@pdfme/common';
+import { replacePlaceholders } from '@pdfme/common';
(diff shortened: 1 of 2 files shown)


hand-dot:
Hi @roopantj !

Thank you for creating the PR! I'll check it and get back to you when I have time. Please wait a bit.

hand-dot:
This PR cannot be merged. While expressions can indeed be used, they do not work for the scenario I want to achieve.

The scenario I want to implement is similar to an Invoice template, where specific columns are set to readOnly, and when a row is added via a form, the values are automatically calculated.

CleanShot 2025-02-03 at 17 59 52

PR #699, opened 2025-02-08
Improve UI font caching and simplify fontKitFont use

### LINKED ISSUE #693: The font cache seems to not be working, which causes slow rendering in Designer, Form and Viewer.
### Describe the bug

Hi @hand-dot, I think I found a serious rendering performance issue. 

The `_cache` used in the following function seems to be almost always empty. So it calls `fetch` to grab the external fonts when rendering text schemas.

https://github.com/pdfme/pdfme/blob/a914cf61fac2475b5fe6d7bf245b2bdd7c010853/packages/schemas/src/text/helper.ts#L110-L135

### To Reproduce

Open the following template, please replace fontName properly, from Designer or Viewer.

https://github.com/pdfme/pdfme/blob/main/packages/generator/__tests__/assets/templates/%E5%B1%A5%E6%AD%B4%E6%9B%B8.json

You can notice that texts are rendered very slowly. In Viewer, if you resize the window, the texts become blank once, then their rerendering start slowly again. Resizing multiple times is disastrous.

https://github.com/user-attachments/assets/11489bef-06b1-442a-964e-c3ac2600b002

### Expected behavior

`NotoSansJP-Regular` has about 6MB size, even though it is cached in the browser, the `fetch` is slower than memory access. It should be cached in the memory as expected.

### Your Environment

```markdown
- pdfme version:5.3.1
- Operating system: macOS Sequoia Version 15.11
- Node.js version or Browser name & version: Version 132.0.6834.160 (Official Build) (arm64)
```

### Your Error Log

```shell
If you resize the window multiple times really quickly, `fetch` fails with the following errors.
 

GET https://fonts.gstatic.com/s/notoserifjp/v30/xn71YHs72GKoTvER4Gn3b5eMRtWGkp6o7MjQ2bwxOubAILO5wBCU.ttf net::ERR_INSUFFICIENT_RESOURCES


Uncaught (in promise) TypeError: Failed to fetch
    at getFontKitFont (index-fa1b6af1.js:1454:326136)
    at uiRender$3 (index-fa1b6af1.js:1577:88)
    at index-fa1b6af1.js:966:208
    at Ii (index-fa1b6af1.js:122:81961)
    at Vj (index-fa1b6af1.js:124:6766)
    at dt.unstable_runWithPriority (index-fa1b6af1.js:114:4215)
    at cg (index-fa1b6af1.js:122:39455)
    at Dj (index-fa1b6af1.js:124:6504)
    at index-fa1b6af1.js:124:6417
    at Tr (index-fa1b6af1.js:114:3068)
```

### Additional context

`_cache` Map is created with an `empty` Map at this line.

https://github.com/pdfme/pdfme/blob/a914cf61fac2475b5fe6d7bf245b2bdd7c010853/packages/ui/src/components/Renderer.tsx#L99

shouldn't  it be stored somewhere more global context?

Since I have almost no experience in ReactJS development, this modification is a burden for me.

peteward:
fixes #693 

Having a `CacheContext` means the cache remains global, not just one cache per schema that is sometimes lost on browser resize.

Also reduced the number of calls to `getFontKitFont()` by pulling it higher up, which means less arguments need passing and less async functions. This is hopefully a bit cleaner.

### DIFF
--- a/packages/schemas/__tests__/text.test.ts
+++ b/packages/schemas/__tests__/text.test.ts
@@ -154,12 +154,15 @@ describe('getSplittedLines test with real font width calculations', () => {
 });
 
 describe('calculateDynamicFontSize with Default font', () => {
-  const font = getDefaultFont();
-  const _cache = new Map();
+  let fontKitFont: FontKitFont;
+
+  beforeAll(async () => {
+    fontKitFont = await getFontKitFont('SauceHanSansJP', getDefaultFont(), new Map());
+  });
 
   it('should return default font size when dynamicFontSizeSetting is not provided', async () => {
     const textSchema = getTextSchema();
-    const result = await calculateDynamicFontSize({ textSchema, font, value: 'test', _cache });
+    const result = calculateDynamicFontSize({ textSchema, fontKitFont, value: 'test' });
 
     expect(result).toBe(14);
   });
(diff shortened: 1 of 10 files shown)


hand-dot (approved):
LGTM!

hand-dot:
Released PDFME_VERSION with '5.3.4'

PR #722, opened 2025-02-19
Add separate build for node

necessarylion:
For the Node.js framework utilizing the NodeNext module, our package encounters the following error:
 
`yntaxError: The requested module '@pdfme/generator' does not provide an export named 'generate'`

To address this issue, I have added a dedicated build script specifically for Node.js applications using the NodeNext module. Typically, Node.js applications require the generator, common, and schema packages. Therefore, I have included these three packages in the build script.

## Summary by CodeRabbit

- **New Features**
  - Enhanced module resolution with dedicated entry points for both Node.js and standard environments.
  - Expanded the build process to include a Node.js-specific step running in parallel with existing builds.
  - Introduced tailored build configurations that optimize TypeScript compilation for Node.js compatibility.

### DIFF
--- a/packages/common/package.json
+++ b/packages/common/package.json
@@ -26,7 +26,10 @@
   "types": "dist/types/src/index.d.ts",
   "exports": {
     ".": {
-      "import": "./dist/esm/src/index.js",
+      "import": {
+        "node": "./dist/node/src/index.js",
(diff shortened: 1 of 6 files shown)


hand-dot (approved):
LGTM! Thank you @necessarylion !!

PR #844, opened 2025-03-11
Available dynamic fields

ianbaek:
This Pull Request adds support for displaying available dynamic fields within the UI, making it easier to both add new fields and see where they’re in use. These fields can be set in Designer’s options as shown below:

```
designer.current = new Designer({
  domContainer: designerRef.current,
  template,
  options: {
    lang,
    font,
    labels: { 'clear': '🗑️' },
    theme: { token: { colorPrimary: '#25c2a0' } },
    availableDynamicFields: ['barcode', 'image', 'company_name', 'company_address'],
  },
  plugins: getPlugins(),
});
```

By providing the _availableDynamicFields_ array directly, the Designer UI will display these fields for users to select and integrate wherever needed.


hand-dot:
Hi @ianbaek !! 

First of all, thank you for your contribution!
This is an interesting feature addition, but it cannot be merged directly into the main branch.

I prefer to keep the library as lightweight as possible, so this kind of processing should be handled outside the library.

In fact, I plan to remove the "Bulk update field names" feature in the future and have it handled outside the library.

This is to reduce the work of writing documentation to explain how to use the library and to minimize maintenance efforts for features that are not widely used.

PR #854, opened 2025-03-14
Restore NotoSerifJP and NotoSansJP in Playground Font Settings

hitomi-t260g:
### Summary  
This PR restores **NotoSerifJP** and **NotoSansJP** in the font settings of the playground by uncommenting their values.  

### Related Discussion  
This issue was discussed in [GitHub Discussions #845](https://github.com/pdfme/pdfme/discussions/845).  

### Why This Change Is Important  
Restoring Japanese fonts makes it easier to confirm that **pdfme supports non-English fonts**, which is valuable for library selection.  
In my case, I used the playground to **demonstrate pdfme’s usability** to our sales team.  
After deciding to adopt pdfme and starting development, we noticed that Japanese fonts were missing, which caused confusion.  

### What I Did  
- Uncommented **NotoSerifJP** and **NotoSansJP** in the playground’s font settings.  

### Verification  
- Confirmed that this change **does not affect e2e tests**.  

This is a simple fix, but it improves usability for users who need **multilingual PDF support**.  
Thank you for your time and consideration! 😊

### DIFF
--- a/playground/src/helper.ts
+++ b/playground/src/helper.ts
@@ -12,14 +12,14 @@ export function fromKebabCase(str: string): string {
 
 export const getFontsData = (): Font => ({
   ...getDefaultFont(),
-  // NotoSerifJP: {
-  //   fallback: false,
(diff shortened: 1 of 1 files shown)


hand-dot:
@hitomi-t260g Okay!

hand-dot:
Deployed
CleanShot 2025-03-14 at 21 23 39

PR #908, opened 2025-04-05
Add z-fold-brochure template

hitomi-t260g:
### 🔄 Add Z-Fold Brochure Template

This PR adds a **Z-Fold Brochure** template to the template collection.

#### 🧾 Overview

The Z-Fold Brochure template allows users to easily create double-sided brochures using a two-page A4 horizontal layout. Each panel is organized to align correctly when folded, including support for elements like maps, product lists, and menus.
It is also a template that promotes understanding of reflecting the number of pages in the basePDF.

- **Format**: Two A4 horizontal pages (double-sided printing)
- **Layout**: Z-fold (tri-fold) with accurate frame alignment
- **Customizable**: Text and images can be replaced using field names
- **Use Case**: Perfect for restaurant menus, store catalogs, event guides, and more

#### 🛠️ How It Works

You can generate multiple variations of the brochure by dynamically replacing content fields in the template.

Instructions on how to edit the template are described in the template.
- Use two A4 horizontal pages when replacing BasePDF.
- Place text and images according to the frame of the BasePDF.
- Double-sided printing is assumed.

Let me know if further refinements or test assets are needed! Looking forward to your review 🙌


hand-dot (approved):
Thanks!!❤️

PR #965, opened 2025-05-02
onChange should be fired only if the schema is in the current page

### LINKED ISSUE #964: A template with tables not on the first page causes an infinite loop
### Describe the bug

This is a very rare edge case, but as mentioned here:

https://discord.com/channels/933602537808990268/933602537808990271/1367183959372468304

If a template contains tables on the second or subsequent pages, pdfme Designer enters an infinite loop at the following lines:

https://github.com/pdfme/pdfme/blob/34e9f4956912c65a16bc5c7b7f7e921f58281206/packages/schemas/src/tables/uiRender.ts#L431-L434

### To Reproduce

Open the following template:

[template.json](https://github.com/user-attachments/files/20012825/template.json)

or to generate such template with:

1. Add a new page
2. Place **two** tables on the second page
3. Save the template
4. Edit the height of the tables in your editor
5. Reload the template
6. Boom! the browser page hangs, all you can do is force close it

The interesting point here is that simply adding one table would not surface the issue, the browser somehow overcome the infinite loop (possibly related related to the animation frame, but not sure). 

### Expected behavior

pdfme should be able to load such a template. The initial table height may differ from the actual height if the template is generated outside of the Designer.

### Your Environment

```markdown
- pdfme version: 5.3.15
- Operating system: 
```

### Your Error Log

```shell
N/A
```

### Additional context

I think I found a temporaly fix for this issue and would post the PR later.

kyasu1:
This is a workaround with my random thought, but it fixes #964. 

If the `schema.height` differs from the calculated `tableHeight`, `onChange` will be triggered at the following code:

https://github.com/pdfme/pdfme/blob/34e9f4956912c65a16bc5c7b7f7e921f58281206/packages/schemas/src/tables/uiRender.ts#L431-L434

which will be handled by the follwoing function:

https://github.com/pdfme/pdfme/blob/34e9f4956912c65a16bc5c7b7f7e921f58281206/packages/ui/src/helper.ts#L567-L591

In this code, on the initial load,`tgt` will be undefined if the schema is not on the first page, since the passed `schemas` argument references the `SchemaForUI`  array of the page zero. This prevents updating `schema.height` and casuses an infinite loop.

My idea is simply to stop passing `onChange` function if the schema is not on the current page at the `Canvas` level.

### DIFF
--- a/packages/ui/src/components/Designer/Canvas/index.tsx
+++ b/packages/ui/src/components/Designer/Canvas/index.tsx
@@ -506,6 +506,18 @@ const Canvas = (props: Props, ref: Ref<HTMLDivElement>) => {
             value = replacePlaceholders({ content, variables, schemas: schemasList });
           }
 
+          const onChange =
+            schemasList[pageCursor].filter((s) => s.id === schema.id).length > 0
+              ? (arg) => {
+                  // Use type assertion to safely handle the argument
+                  type ChangeArg = { key: string; value: unknown };
+                  const args = Array.isArray(arg) ? (arg as ChangeArg[]) : [arg as ChangeArg];
+                  changeSchemas(
+                    args.map(({ key, value }) => ({ key, value, schemaId: schema.id })),
+                  );
+                }
+              : undefined;
+
           return (
             <Renderer
               key={schema.id}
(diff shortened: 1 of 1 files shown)


hand-dot:
> The height of the table schemas that are not on the first page will not be updated unless you double-click them or put them into edit mode. 

Doesn't this behavior cause issues for users who are using the viewer?
If the table on the second page is rendered correctly, then it's not a problem, though...

kyasu1:
As far as I have tested, I have not experienced any unexpected behavior on the Form/Viewer.

Issue #953, opened 2025-04-25
Adding a text schema should default the font to either the first or your default font


peteward:
### Is your feature request related to a problem? Please describe.

For every text schema that you add to the page you need to explicitly set the font for it. 

Whilst the text schema widget does provide the fontName as a placeholder, it doesn't actually save this value as it is not set. This means that while editing you might think that you have selected 'Roboto' as a font because this is what is displayed, but actually no font is set. This means that when you render you are relying on the default font in the UI and the default font in the generator matching.  

In our case we are seeing templates created months ago and the default font changing when rendered weeks or months later.

### Describe the solution you'd like

Default the font of every next text field added to the provided font with fallback=true set (or the first in the list if none have this)

### Describe alternatives you've considered

-

PR #999, opened 2025-05-11
Allow defaultSchema to be a function to consider Options

peteward:
implements #953 

Allowing the default schema to be specified by a `defaultSchemaFn` which can have options passed to it. The props only have UI Options for now, but could be extended in future if other schemas require it.

This will allow users to customise what the defaults are for any new schema based on what's passed. In this case it's the `fontName` from the passed in fonts.


packages/common/src/types.ts
```diff
@@ -118,20 +118,31 @@ type PropPanelProps = {
...
+ * @property {(props: DefaultSchemaProps) => T)} defaultSchemaFn - An optional function to generate the base schema dynamically using run-time args
  */
 export interface PropPanel<T extends Schema> {
   schema:
     | ((propPanelProps: Omit<PropPanelProps, 'rootElement'>) => Record<string, PropPanelSchema>)
     | Record<string, PropPanelSchema>;
-
   widgets?: Record<string, (props: PropPanelWidgetProps) => void>;
   defaultSchema: T;
+  defaultSchemaFn?: (props: DefaultSchemaProps) => T;
```
peteward, line 145: I couldn't think of a better way of supporting a function than this...

hand-dot: How about this?

```ts
  defaultSchema: 
  | T 
  | ((props: DefaultSchemaProps) => T);
```

peteward: Yeah we could do that.

There are quite a lot of places in the code that use the `defaultSchema` that would need to be updated to check if it's a function though, not just the one place I changed in the LeftSidebar.

peteward: i'll try 👍

peteward: I tried this:

```
  if (schema.fontName == undefined) {
    onChange && onChange({ key: 'fontName', value: getFallbackFontName(font) });
  }
```

it works for text schema, but for MVT and tables it causes infinite loop.

I think the previous suggestion will be simpler than I thought, I will try it

peteward:
closing for a simpler solution

PR #1004, opened 2025-05-12
feat: Add repeatHead option to table schema and UI translations

SHUHAIB-T:
# Repeat Header Feature Documentation

## Overview

The "Repeat Header" feature enhances PDF templates by allowing table headers to be repeated on each page when content spans multiple pages. This document explains the implementation and usage of this feature.

## Problem Statement

In multi-page PDF documents with tables, by default, table headers appear only on the first page. This can make reading tables that span multiple pages difficult and confusing.

## Solution

The "Repeat Header" feature allows users to specify that table headers should repeat on every page when a table spans multiple pages.

## Implementation

The feature was implemented by modifying the existing header visibility logic:

```javascript
  const copyOfShowHead = schema.showHead;
  schema.showHead = copyOfShowHead === false ? false : !schema.__isSplit;

  // this will make sure the head is shown when the table is split and repeatHead is true
  if (schema.repeatHead && schema.__isSplit && copyOfShowHead !== false) {
    schema.showHead = true;
  }

```

### Logic Explanation

1. First, we save the original value of `showHead` to preserve user intention
2. Next, we apply the existing logic: if `showHead` is explicitly `false`, keep it `false`; otherwise, show the header only on the first page (when `__isSplit` is `false`)
3. Finally, we implement the repeat header feature: if `repeatHead` is `true`, content spans multiple pages (`__isSplit` is `true`), and the original `showHead` wasn't explicitly `false`, then we set `showHead` to `true` to ensure headers appear on subsequent pages

This implementation maintains backward compatibility while adding the new feature.

### Enabling Repeat Headers

To enable header repetition across pages:

1. Open the template editor
2. Select the table component
3. In the properties panel, find the "Repeat Header" option
4. Toggle the switch to enable the feature
screenshort_1

#### Without Repeat Header (Default)
- Table headers appear only on the first page
- Subsequent pages show only data rows

#### With Repeat Header Enabled
- Table headers appear on every page where the table continues
- Improves readability for multi-page tables

image

#### Priority Rules

The feature respects these priority rules:

1. If `showHead` is explicitly set to `false`, no headers will show (regardless of `repeatHead`)
2. If `showHead` is not `false` and `repeatHead` is `true`, headers will appear on all pages
3. If `repeatHead` is not specified, the default behavior applies (headers only on first page)

## Configuration Example

```json
{
  "schema": {
    "showHead": true,
    "repeatHead": true,
    // other schema properties...
  }
}
```

## Benefits

- Improved readability for multi-page tables
- Better usability for long documents
- Consistent document style
- Professional-looking outputs

---


hand-dot:
@SHUHAIB-T 

Thank you.
After a page break, we need to account for the header height in elements rendered below the table.

https://github.com/user-attachments/assets/dafec458-9bff-40ac-b0ab-6ede0009c5a3

SHUHAIB-T:
@hand-dot , this is how I am calculating the header height. 

```typescript

function getTableHeaderHeight(schema: any): number {
  const headStyles = schema?.headStyles;
  const lineHeight = headStyles?.lineHeight || 1;
  const paddingTop = headStyles.padding?.top || 0;
  const paddingBottom = headStyles.padding?.bottom || 0;
  const fontSize = headStyles?.fontSize || 0;
  const headerHeight = fontSize + paddingTop - lineHeight - paddingBottom;
  return headerHeight;
}
```

1- Is there an existing method in the library (or a recommended approach) to calculate this, rather than manually computing it?

2- How can I improve this function for better accuracy?

Any guidance would be appreciated!

hand-dot:
> 2- How can I improve this function for better accuracy?

You can use [heightOfFontAtSize](https://github.com/pdfme/pdfme/blob/main/packages/schemas/src/text/helper.ts#L74) to calculate text height.

SHUHAIB-T:
@hand-dot , I have moved the methods getFontKitFont and HeightofFontAtSize to common as will be required for all packages.

SHUHAIB-T:
@hand-dot ,  I've improved the `getTableHeaderHeight` function to calculate the height more accurately, considering line height, padding, and border width.

[Screencast](https://github.com/user-attachments/assets/97a7d7d2-6b4d-4966-a241-b943f1801d25)

hand-dot:
@SHUHAIB-T cool, I will review this today, thanks man

hand-dot:
Hi @SHUHAIB-T 

I built and tested the code locally.
When performing a page break to the 3rd or 4th page using the Invoice or Quotes templates, the layout breaks as shown in the GIF below.

CleanShot 2025-05-14 at 15 26 55


I will review the code and also think about how to fix this issue, but unless this problem is resolved, I won’t be able to proceed with the merge.

SHUHAIB-T:
@hand-dot, I will check this.

hand-dot:
Handled by https://github.com/pdfme/pdfme/pull/1025

PR #1005, opened 2025-05-12
Avoid full schema rerender while dragging new field

peteward:
I noticed that when dragging a new schema out from a panel icon that the schema was being fully re-rendered for every pixel movement of the mouse.

This memoizes so that it does not rerender again until dropped.


hand-dot (approved):
LGTM!

PR #1007, opened 2025-05-15
improve DX with pluginRepository and make plugin checking more robust

peteward:
There is a lot of unnecessary type checking and complexity around handling plugins.

This PR wraps the existing `Plugins` object in some convenience methods for simpler use and improves visibility of type errors

### DIFF
--- a/packages/common/src/pluginRegistry.ts
+++ b/packages/common/src/pluginRegistry.ts
@@ -6,10 +6,10 @@ import { Plugins, Plugin, PluginRegistry } from './types.js';
 export const pluginRegistry = (plugins: Plugins): PluginRegistry => {
   return {
     plugins: plugins,
-    entries: (): [string, Plugin<any>][] => Object.entries(plugins),
-    values: (): Plugin<any>[] => Object.values(plugins),
+    entries: (): [string, Plugin][] => Object.entries(plugins),
+    values: (): Plugin[] => Object.values(plugins),
     exists: (): boolean => Object.values(plugins).length > 0,
-    findWithLabelByType(type: string): [string, Plugin<any> | undefined] {
+    findWithLabelByType(type: string): [string, Plugin | undefined] {
       for (const [label, plugin] of Object.entries(this.plugins) as [string, Plugin][]) {
         if (!plugin || typeof plugin !== 'object') continue;
         if (!plugin.propPanel || typeof plugin.propPanel !== 'object') continue;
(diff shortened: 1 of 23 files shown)


packages/common/src/schema.ts
```diff
@@ -152,12 +152,25 @@ export const Font = z.record(
...
+    icon: z.string().optional(),
+  })
+  .passthrough();
+
 export const CommonOptions = z.object({ font: Font.optional() }).passthrough();
 
 const CommonProps = z.object({
   template: Template,
   options: CommonOptions.optional(),
-  plugins: z.record(z.object({ ui: z.any(), pdf: z.any(), propPanel: z.any() })).optional(),
```
peteward, line 160: the problem with `z.any()` is that even the existence of the key is not checked, so doesn't offer any value.
I have used a more detaile definition of `Plugin` above and used `z.unknown()` for similar flexibility to `z.any()` but as least the key is checked with `parse()`

packages/common/src/types.ts
```diff
@@ -153,8 +153,16 @@ export type Plugin<T = Schema> = {
   uninterruptedEditMode?: boolean;
 };
 
-// eslint-disable-next-line @typescript-eslint/no-explicit-any
-export type Plugins = { [key: string]: Plugin<any> | undefined };
+export type Plugins = { [key: string]: Plugin<any> };
```
peteward, line 156: I couldn't see a reason why an element of a the plugins object could be `undefined`?

packages/generator/src/helper.ts
```diff
@@ -80,11 +80,9 @@ export const preprocessing = async (arg: { template: Template; userPlugins: Plug
...
   pdfDoc.registerFontkit(fontkit);
 
-  const pluginValues = (
-    Object.values(userPlugins).length > 0
-      ? Object.values(userPlugins)
-      : Object.values(builtInPlugins)
-  ) as Plugin<Schema>[];
+  const plugins = pluginRegistry(
+    Object.values(userPlugins).length > 0 ? userPlugins : builtInPlugins,
+  );
```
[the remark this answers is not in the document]
peteward: preprocessing is only called once per generate...

packages/common/src/pluginRegistry.ts
```diff
@@ -0,0 +1,32 @@
+import { Plugins, Plugin, PluginRegistry } from './types.js';
+
+/**
+ * Wraps plugins collection with utility methods
+ */
+export const pluginRegistry = (plugins: Plugins): PluginRegistry => {
+  return {
+    plugins: plugins,
+    entries: (): [string, Plugin<any>][] => Object.entries(plugins),
```
peteward, line 9: need to improve on `<any>`

hand-dot (approved):
Coooool!

PR #1010, opened 2025-05-16
feat: Add invoice-ja-simple Template

EedgeY:
This is an invoice template designed for Japanese users.
It is a basic and minimal version, intended as the simplest template.


hand-dot (approved):
🫶

Issue #1008, opened 2025-05-15
MVT loses focus when typing '}' in editor mode


peteward:
### Describe the bug

As soon as you type a closing bracket this now loses focus - it didn't use to do this so needs checking when this regression was introduced

### To Reproduce

add MVT to page
start typing a variable with `{xyz`
as soon as you type `}` focus will be lost.

### Expected behavior

You should continue typing until you choose to end focus

### Your Environment

```markdown
- pdfme package(@pdfme/generator or @pdfme/ui): ui
- pdfme version: latest main
```

### Your Error Log

```shell
-
```

PR #1011, opened 2025-05-16
Multivariable text fixes

### LINKED ISSUE #1009: MVT throws an SyntaxError when removing previously typed fields for a moment
### Describe the bug

A SyntaxError is thrown briefly when trying to parse a non-JSON string to JSON:

```
Unexpected token 'i', "first name" is not valid JSON
```

This appears to apply a variable name from the previous list of JSON variables.

### To Reproduce

* Create an MVT field
* Type some content with 1 or more variables
* lose focus on the field
* edit the field again, use the mouse to select all text then replace it with something without variables
* lose focus

### Expected behavior

No JSON error should be thrown

### Your Environment

```markdown
- pdfme version:
- Operating system:
```

### Your Error Log

```shell
SyntaxError: Unexpected token 'a', "names" is not valid JSON
    at JSON.parse (<anonymous>)
    at substituteVariables (helper.js?t=1747379785557:11:24)
    at uiRender (uiRender.js?t=1747379785998:15:50)
    at index.es.js:153217:12
    at Zc (index.es.js:3529:24)
    at Sf (index.es.js:4365:24)
    at a.unstable_runWithPriority (index.es.js:551:16)
    at Xn (index.es.js:2264:23)
    at Da (index.es.js:4349:23)
    at Ms (index.es.js:3991:9)
```

### Additional context

This appears to be due to the following code in Renderer.tsx:

```
          if (mode !== 'designer' && schema.readOnly) {
            const variables = {
              ...schemasList.flat().reduce(
                (acc, currSchema) => {
                  acc[currSchema.name] = currSchema.content || '';
                  return acc;
                },
                {} as Record<string, string>,
              ),
              totalPages: schemasList.length,
              currentPage: index + 1,
            };

            value = replacePlaceholders({ content, variables, schemas: schemasList });
          }
```

An MVT field should only be `readOnly` when it has no variables, so normally this isn't a problem. It seems that due to the sequence of changes the readOnly is set briefly before the variables are emptied - leaving placeholders that this code picks up and tries to replace.

peteward:
fixes #1009 #1008

### DIFF
--- a/packages/schemas/src/multiVariableText/propPanel.ts
+++ b/packages/schemas/src/multiVariableText/propPanel.ts
@@ -15,6 +15,7 @@ const mapDynamicVariables = (props: PropPanelWidgetProps) => {
     changeSchemas([
       { key: 'content', value: JSON.stringify(variables), schemaId: activeSchema.id },
       { key: 'variables', value: varNames, schemaId: activeSchema.id },
+      { key: 'readOnly', value: varNames.length === 0, schemaId: activeSchema.id },
     ]);
   }
 
--- a/packages/ui/src/components/Renderer.tsx
+++ b/packages/ui/src/components/Renderer.tsx
@@ -28,7 +28,7 @@ type RendererProps = Omit<
 };
 
 type ReRenderCheckProps = {
-  plugin?: Plugin<Schema>;
+  plugin: Plugin<Schema>;
   value: string;
   mode: Mode;
   scale: number;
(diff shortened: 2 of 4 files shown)


packages/schemas/src/multiVariableText/propPanel.ts
```diff
@@ -15,6 +15,7 @@ const mapDynamicVariables = (props: PropPanelWidgetProps) => {
     changeSchemas([
       { key: 'content', value: JSON.stringify(variables), schemaId: activeSchema.id },
       { key: 'variables', value: varNames, schemaId: activeSchema.id },
+      { key: 'readOnly', value: varNames.length === 0, schemaId: activeSchema.id },
```
peteward, line 18: setting readOnly here, at the same time as the variables and content works around the #1009 problem

packages/ui/src/components/Renderer.tsx
```diff
@@ -147,13 +153,7 @@ const Renderer = (props: RendererProps) => {
         ref.current.innerHTML = '';
       }
     };
-  }, [plugin?.ui, schema.type, reRenderDependencies]);
```
peteward, line 150: the main bug here, introduced by devin, was that `reRenderDependencies` is already an array and nesting it didn't work properly

the other checks are not necessary because we've move a check for `plugin` and `plugin.ui` further up so they must exist (along with `schema.type`)

packages/ui/src/components/Renderer.tsx
```diff
@@ -147,9 +147,9 @@ const Renderer = (props: RendererProps) => {
...
+  }, reRenderDependencies);
```
peteward, line 150: the main problem was here that `reRenderDependencies` is already an array and nesting it seemed to break this

hand-dot (approved):
LGTM

PR #1013, opened 2025-05-16
Add Invoice ja simple landscape template

EedgeY:
Simple invoice template (landscape format) for Japanese users.


hand-dot (approved):
LGTM!

PR #1015, opened 2025-05-17
Default fontName to fallback font for all newly created schemas in UI

### LINKED ISSUE #953: Adding a text schema should default the font to either the first or your default font

peteward:
fixes #953

This is the simplest solution that I can think of. It's not perfect if people use a different key from `fontName` in their schemas, but it works for all current use cases, including nested text within tables.


hand-dot (approved):
LGTM!

PR #1052, opened 2025-06-04
Export blob2Base64Pdf function

tylerbarker:
Hey again, this is a very handy function and I noticed I couldn't import it from `@pdfme/common` - reckon we can export it?


hand-dot:
Hi @tylerbarker !

No, I don't want to export this method because it only works in the browser 🙇‍♂️
Please use getB64BasePdf instead.

Issue #484, opened 2024-06-03
Table capabilities


romankolosok:
### Is your feature request related to a problem? Please describe.

I want to have capability to use different fonts within a table, for example I need to create a table, where some fields are just text, and other fields are barcodes to that item in that row. I want to be able to change font in the barcode cell to an actual barcode39/128 font.

### Describe the solution you'd like

I would like to be able to have an option to select a custom font for each cell in the table, to be able to generate auto table where each row contains text of different fonts.

### Describe alternatives you've considered

I considered dynamically creating however many barcodes and putting them near the corresponding table cell but encountered issue with dynamic table, if my text is longer than the column width, it wraps around and increases height of the row and I cannot determine height of the row before rendering the pdf, so my barcodes are not mapping to the rows they are corresponding to.

hand-dot:
@romankolosok 
Are you saying that you want to place barcodes within table cells?

romankolosok:
Yes, I had to work around and "draw" table manually with rectangles and static height to put barcodes in it. And it significantly affects pdf generation time I'm pretty sure, I think it would be very helpful If I could embed barcode within a table cell.

romankolosok:
I tested it on the biggest data feed I had: around 900 rows of data, It took around 4 minutes to generate

hand-dot:
I'd like to treat performance issues as a separate problem for now. While I understand that the nested for loops are degrading performance, there are still bugs in the table itself.

I think it's a good suggestion to allow specifying schemas other than text in cells, but we're not at a stage where we can incorporate that suggestion yet. However, I'm not rejecting this proposal, so I'd like to change the title to something more appropriate.

We might be able to implement this once the table bugs and performance issues are resolved.

brian-scb:
👍 also looking for similar. Nested tables would be awesome.

joernroeder:
images or custom schemas inside of cells would be awesome too

nicofishman:
Yeah, it would be awesome to be able to put checkboxes inside of table cells

hand-dot:
ref: https://github.com/pdfme/pdfme/pull/1241/files

superiority-claw-agent:
[...]

PR #1025, opened 2025-05-31
Add repeatHead option to TableSchema

hand-dot:
Introduce a new `repeatHead` property to the `TableSchema` interface, allowing for the header to be repeated in split tables. Update relevant components and translations accordingly.

---

ref: #629

Issue #721, opened 2025-02-17
Support Different Font Styles per Column in Tables


jeandepeyr:
### Is your feature request related to a problem? Please describe.

Currently, it seems that table body columns share the same font style. It would be useful to allow different styles per column or even inline styling within table cells.

### Describe the solution you'd like

Ideally, having the option to define a custom schema per row could offer more flexibility in table design.
What do you think ?

### Describe alternatives you've considered

The only thing we can do right now is giving a different style for the header or the body. We can not have only the last column in bold font for instance.

baliya367:
I may have found a solution to your problem. You can look here [click](https://javierlopez.eu/captha/) Hope this helped

hand-dot:
Closing as covered by #484. Per-column/per-cell font styling is part of the broader table-cell rendering request, which also covers non-text/custom schema content inside table cells.

PR #1055, opened 2025-06-09
feat: Enhance table styling to support row and cell-level custom styles

smiddleton93:
### Summary

The purpose of this PR is to enhance the granular control over the styling of columns, rows and individual cells within the table schema. 

**Current functionality:**
- Styling headers and body sections
- Applying alignment or font name per column

**New functionality:**
- Apply styling to specific rows
- Apply styling to specific cells within a row
- Expand the styling that can be applied to specific columns

### Example usage
```
const robotoRegularData = readFileSync(path.join(__dirname, `/assets/fonts/Roboto-Regular.ttf`));
const robotoMediumData = readFileSync(path.join(__dirname, `/assets/fonts/Roboto-Medium.ttf`));

const font: Font = {
  default: { fallback: true, data: robotoRegularData },
  bold: { data: robotoMediumData },
};

const plugins = {
  text,
  table,
};
const rows = [
  ['Row 1 Col 1', 'Row 1 Col 2'],
  ['Row 2 Col 1', 'Row 2 Col 2'],
  ['Row 3 Col 1', 'Row 3 Col 2'],
  ['Row 4 Col 1', 'Row 4 Col 2'],
  ['Row 5 Col 1', 'Row 5 Col 2'],
  ['Row 6 Col 1', 'Row 6 Col 2'],
];

const content = JSON.stringify(rows);

const tableSchema: TableSchema = {
  name: `table-schema`,
  type: 'table',
  position: { x: 15, y: 15 },
  width: 180,
  height: 150,
  content,
  showHead: true,
  head: ['Header 1', 'Header 2'],
  headWidthPercentages: [50, 50],
  tableStyles: {
    borderWidth: 0.3,
    borderColor: '#888888',
  },
  headStyles: {
    fontName: 'bold',
    fontSize: 10,
    characterSpacing: 0,
    alignment: 'left',
    verticalAlignment: 'middle',
    lineHeight: 1,
    fontColor: '#ffffff',
    borderColor: '',
    backgroundColor: '#2980ba',
    borderWidth: {
      top: 0,
      right: 0,
      bottom: 0,
      left: 0,
    },
    padding: {
      top: 5,
      right: 5,
      bottom: 5,
      left: 5,
    },
  },
  bodyStyles: {
    fontName: 'Roboto',
    fontSize: 10,
    characterSpacing: 0,
    alignment: 'left',
    verticalAlignment: 'middle',
    lineHeight: 1,
    fontColor: '#ffffff',
    borderColor: '#888888',
    backgroundColor: '#ffffff',
    alternateBackgroundColor: '#ffffff',
    borderWidth: {
      top: 0.1,
      right: 0.1,
      bottom: 0.1,
      left: 0.1,
    },
    padding: {
      top: 2,
      right: 5,
      bottom: 2,
      left: 5,
    },
  },
  rowStyles: {
    0: {
      backgroundColor: '#34ebd8',
    },
    2: {
      backgroundColor: '#ffff00',
    },
    4: {
      cells: {
        1: { backgroundColor: '#ff0000' },
      },
    },
  },
  columnStyles: {
    fontName: {
      0: 'bold',
    },
    textColor: {
      0: '#00ff00',
      1: '#0000ff',
    },
    alignment: {
      0: 'left',
      1: 'right',
    },
  },
  required: false,
  readOnly: true,
};

const template: Template = {
  basePdf: { width: 210, height: 297, padding: [10, 10, 10, 10] },
  schemas: [[tableSchema]],
};
```

image

## Issue linkage

Related to #484. Row/cell-level styles can support the per-cell font styling part of the request. Embedding barcode/image schema objects in table cells and sizing rows around them are outside this PR, so the broader non-text-cell issue should remain open.

### DIFF
--- a/packages/schemas/src/tables/tableHelper.ts
+++ b/packages/schemas/src/tables/tableHelper.ts
@@ -207,6 +207,29 @@ function mapColumnStyles(schema: TableSchema): Record<number, Partial<Styles>> {
   }, columnStylesWidth);
 }
 
+function adjustRowStyles(
+  rowStyles:
+    | { [rowIndex: number]: Partial<Styles> & { cells?: { [colIndex: number]: Partial<Styles> } } }
+    | undefined,
+  __bodyRange: { start: number; end?: number },
+):
+  | { [rowIndex: number]: Partial<Styles> & { cells?: { [colIndex: number]: Partial<Styles> } } }
+  | undefined {
+  if (!rowStyles || __bodyRange.start === 0) return rowStyles;
+
+  const adjusted: typeof rowStyles = {};
+
+  for (const key in rowStyles) {
+    const originalIndex = Number(key);
+    const adjustedIndex = originalIndex - __bodyRange.start;
+    if (adjustedIndex >= 0) {
+      adjusted[adjustedIndex] = rowStyles[originalIndex];
+    }
+  }
+
+  return adjusted;
+}
+
 function getTableOptions(schema: TableSchema, body: string[][]): UserOptions {
   return {
     head: [schema.head],
@@ -217,7 +240,8 @@ function getTableOptions(schema: TableSchema, body: string[][]): UserOptions {
     tableLineColor: schema.tableStyles.borderColor,
     tableLineWidth: schema.tableStyles.borderWidth,
     headStyles: mapCellStyle(schema.headStyles),
-    rowStyles: schema.rowStyles || {},
+    rowStyles:
+      adjustRowStyles(schema.rowStyles, schema.__bodyRange || { start: 0, end: undefined }) || {},
     bodyStyles: mapCellStyle(schema.bodyStyles),
     alternateRowStyles: { backgroundColor: schema.bodyStyles.alternateBackgroundColor },
     columnStyles: mapColumnStyles(schema),
--- a/packages/schemas/src/tables/tableHelper.ts
+++ b/packages/schemas/src/tables/tableHelper.ts
@@ -133,7 +133,7 @@ function cellStyles(
 
   const otherStyles = Object.assign({}, styles.styles, sectionStyles);
 
-  const colStyles = styles.columnStyles[column.index] || styles.columnStyles[column.index] || {};
+  const colStyles = styles.columnStyles[column.index] || {};
 
   const rowStyles =
     sectionName === 'body'
--- a/packages/schemas/src/tables/tableHelper.ts
+++ b/packages/schemas/src/tables/tableHelper.ts
@@ -19,7 +19,14 @@ import type {
 } from './types.js';
 import { Cell, Column, Row, Table } from './classes.js';
 
-type StyleProp = 'styles' | 'headStyles' | 'bodyStyles' | 'alternateRowStyles' | 'columnStyles';
+type StyleProp =
+  | 'styles'
+  | 'headStyles'
+  | 'bodyStyles'
+  | 'alternateRowStyles'
+  | 'columnStyles'
+  | 'cellStyles'
+  | 'rowStyles';
 
 interface CreateTableArgs {
   schema: Schema;
@@ -37,7 +44,9 @@ interface UserOptions {
   tableLineColor?: string;
   head?: string[][];
   body?: string[][];
-
+  rowStyles: {
+    [key: number]: Partial<Styles> & { cells?: { [colIndex: number]: Partial<Styles> } };
+  };
   styles?: Partial<Styles>;
   bodyStyles?: Partial<Styles>;
   headStyles?: Partial<Styles>;
@@ -121,13 +130,19 @@ function cellStyles(
   } else if (sectionName === 'body') {
     sectionStyles = styles.bodyStyles;
   }
+
   const otherStyles = Object.assign({}, styles.styles, sectionStyles);
 
   const colStyles = styles.columnStyles[column.index] || styles.columnStyles[column.index] || {};
 
   const rowStyles =
-    sectionName === 'body' && rowIndex % 2 === 0
-      ? Object.assign({}, styles.alternateRowStyles)
+    sectionName === 'body'
+      ? Object.assign(
+          {},
+          rowIndex % 2 === 0 ? styles.alternateRowStyles : {},
+          styles.rowStyles?.[rowIndex] || {},
+          styles.cellStyles?.[rowIndex]?.[column.index] || {},
+        )
       : {};
 
   const defaultStyle = {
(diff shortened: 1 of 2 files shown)


smiddleton93:
https://github.com/pdfme/pdfme/issues/721

hand-dot:
Hi @smiddleton93 👋 
Thank you for sending the PR. I think the idea is great.

Isn’t it possible to configure it from the designer using the [propPanel](https://pdfme.com/docs/custom-schemas#overview-of-custom-schemas--plugins) ? Honestly, 
I can’t think of an optimal UI.

I don’t want to create a hidden feature to keep minimize the codebase.
If it’s something only you need, you can use it as a plugin for yourself.

I’d want to see at least 10 other users requesting this feature to consider it.

smiddleton93:
> Hi @smiddleton93 👋 Thank you for sending the PR. I think the idea is great.
> 
> Isn’t it possible to configure it from the designer using the [propPanel](https://pdfme.com/docs/custom-schemas#overview-of-custom-schemas--plugins) ? Honestly, I can’t think of an optimal UI.
> 
> I don’t want to create a hidden feature to keep minimize the codebase. If it’s something only you need, you can use it as a plugin for yourself.
> 
> I’d want to see at least 10 other users requesting this feature to consider it.

Hi @hand-dot , thank you for taking the time to review the PR and for your feedback.

I completely understand wanting to minimise the code base and avoid introducing features without broad demand. Just to provide some context, I wanted to explain a bit more about our use case and why this feature was important for us:

Our primary use of the library is to generate nutritional/health-related reports server-side, where styling requirements are based on designs from our product/UX team. We're not currently using the Designer package for a few reasons:
- We don’t want end users modifying the schema or layout due to legal and compliance restrictions.
- Instead, we generate the PDF on the backend and use the Form View on the frontend to render the PDF with only a few editable fields (like optional notes or comments).

Some of the data fetching and processing involved in these reports is fairly intensive, so backend generation makes the most sense. The generated PDF is then sent to the frontend as the basePdf. Alongside that, the endpoint specifies which fields are editable - those are excluded from the basePdf and used as inputs in Form View.

I’m also not too sure how this feature would best fit into the Designer UI, as I haven’t looked through that part of the codebase thoroughly. My initial thought would be something like enabling styling options when selecting a specific row, column, or cell - but I’m happy to help explore ideas if needed, although my time may be limited as this is a work project.

Could you clarify what you mean by using this as a plugin? I understand there’s support for custom schemas, but part of the reason I wanted to integrate with the current table schema was to take advantage of features like page breaking, especially since we often have multiple dynamic tables that can span atleast 2–3 pages. I wasn’t sure if that would be straightforward to replicate with a custom implementation.

Thanks again for your time and for your contributions to this great library!


Side note: I've been testing my changes with a complex pdf I'm working on and found an issue with styles when the table is split across a page break. I hadn't realised that the row index starts at 0 on the new page (**Edit** - Pushed commit with fix for this just for future reference). I will keep working on this on my fork, I'm also keeping an eye on this [PR you have open](https://github.com/pdfme/pdfme/pull/1025) as I think changes in there will relate to some other page break issues I've come across while working with tables. I'll be working on this project for quite some time so I will look into ideas for how it could integrate with designer, even if you don't want to consider the feature for some time I'm interested in having a play around with it regardless

peteward:
One of the challenges with pdfme at the moment is that although it supports plugins via custom schemas, it doesn't easily allow in-depth extending or modifying existing schemas. This would be the ideal here - to be able to layer your customisations on top of the table schema. 

The table schema is super-complex and there are almost limitless possibilities for modifications to suit different use cases. It would be good to think about how this could be supported to allow users to customise it more easily without having to fork the project. The standard text schema is similar, developers might want to add their own features to it. We did some work to export some of the functions for re-use with the multiVariableText schema, but the design and pattern could be improved.

hand-dot:
Thank you @smiddleton93 @peteward .

I can understand @peteward is saying.

If we could enable customization of existing schemas, it should allow @smiddleton93 to achieve what they want in a simple way.

However, I can't think of how to implement this. Do you have any good ideas?

hand-dot:
@smiddleton93 
If time is limited, the short-term solution would be to create a fork.
As you can see, there's no clear solution at this point.

However, if you plan to use pdfme long-term, it would be better to use the original library from a maintenance perspective.
If we could build a mechanism that allows customization of existing schemas like the table schema, it would enable the use of the original library.

smiddleton93:
@hand-dot Yes I think I will just be creating a fork for now. I was initially just trying to avoid that, like you said it creates more maintenance for myself and thought someone may be able to make use of the changes I needed to make either way. 
@peteward You make a good point that there could be limitless different customisations and the table schema is complex enough as it is. Some form of middleware pattern, custom render functions or meta properties could work, it would need to be well thought out. I will keep it in mind as I work with the library.

Unrelated question but any suggestions for how to work with developing in the designer area of the ui package? Had a bit of a look through it yesterday, it was a bit difficult to grasp the concepts so I was wondering if there was a better workflow for some visual feedback

hand-dot:
> Unrelated question but any suggestions for how to work with developing in the designer area of the ui package? Had a bit of a look through it yesterday, it was a bit difficult to grasp the concepts so I was wondering if there was a better workflow for some visual feedback

We recommend proceeding with development on the playground using the reference below.
https://github.com/pdfme/pdfme/blob/main/DEVELOPMENT.md

mde:
I need this functionality, and was kind of surprised to see that the column styling consists of precisely one property. :)

I am building a (programmatically generated) digital version of a sign-in sheet for medical compliance education, and need the "signature" column to appear in a cursive font. Being able to style different columns with bolding, italics, etc. seems like pretty basic functionality to me, and the PR doesn't seem to introduce a ton of change from the LOC perspective.

I would love to see this merged so I'm not depending on a person's private fork. (Hi, @smiddleton93! Thank you for doing this. 👍🏻)

joernroeder:
I need this too. cells which have the amounts and sum up values needs different styling apart from text alignment.

joernroeder:
Table footer styles is another example when this would be useful.

nokka:
Hello! I also need this to style different rows with background colors or slight changes to the font such as bold or italic.

nokka:
Hello @hand-dot any chance we can merge this? The code change is minimal and it would provide a lot of benefit. Or if it's better to create a a new plugin for it, I'm not sure what's best. I'm new here :)

hand-dot:
I'm reviewing all the comments.

The current concern is that this setting cannot be changed from the propPanel, so designers cannot modify this setting. This could confuse users who are using the ui package.

For example, if rowStyles is set but you want to remove it, what should users do if they don't want to edit the template JSON (ex: non-programer)?

Also, I wanted to make it the same data format as columnStyles, but was there a specific reason for changing how the data is structured?

JonathanDosmann-pager:
Just adding my voice to this thread. My use case as a programmer who wants to style individual cells conditionally would love to see this merged.

pedroayresb:
Also adding my voice, love the lib, been using it for a while now, and this funcionality would greatly improve my use for it

HugoDaudt:
Love the lib, would love this func.
Suggestion to @hand-dot:
Maybe the lib could improve this func now, without the designer in mind. And after implement it to designer

tbshfr:
This feature would be a huge help for my use case.
I would really appreciate it if it could be merged (even without the designer functionality).
If possible, could we also get some insight into whether this PR is still under consideration in its current form? That would help me decide whether to look for another approach

hand-dot:
Thank you all for your comments. I will consider them carefully. Thank you.

mzex:
This feature will be very important to me, right now I am considering some alternate solution even if I am already using pdfme because of customer requirement to have some data colored in table based on some condition. If this is not going to break anything why not to merge it?

danielwillms:
This feature would be very helpful to me, are there any news on if this will be merged in the near future?

6LWa6ZKx:
Not only that. "Table" is a very useful "container". When designing different styles of printed documents. There is a high demand for merging and splitting operations. It would be great if the selected cells could be dragged to perform the related cell merging operations.

PR #1057, opened 2025-06-17
Bump canvas version for Node 22 support

smiddleton93:
See https://github.com/Automattic/node-canvas/issues/2515 

Currently I'm needing to override this version for environments using node v22

### DIFF
--- a/packages/converter/package.json
+++ b/packages/converter/package.json
@@ -51,10 +51,10 @@
   "devDependencies": {
     "@pdfme/generator": "*",
     "@types/estree": "^1.0.6",
-    "canvas": "^2.11.2"
+    "canvas": "^3.1.0"
   },
   "optionalDependencies": {
-    "canvas": "^2.11.0"
+    "canvas": "^3.1.0"
   },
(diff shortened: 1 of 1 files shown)


hand-dot:
Hi @smiddleton93 !

We are using **node canvas** v2 for the following reasons:

* **pdfjs-dist** only works with v3 series in Node.js → As stated in the [FAQ](https://github.com/mozilla/pdf.js/wiki/Frequently-Asked-Questions#legacy-build), Node.js needs to use the legacy build
* **node canvas** only works with v2 series
   * **pdfjs-dist** v3 series specifies `"canvas": "^2.11.2"` in `optionalDependencies`, so we are aligning with that specification

Please let me know if you have any objections.

smiddleton93:
@hand-dot 
Thank you for clarification. I ran into this issue when using the pdf2img converter in my testing environment for snapshot testing. My testing environment is using node 22 which isn't compatible with v2 without extra steps, by the sounds of it overriding the version is the only solution for now

PR #1159, opened 2025-08-22
schemas(barcodes): add QR (version/mask/qzone) and PDF417 controls; g…

lsadehaan:
…eneric ITF; SVG output with browser fallback; text controls; remove border UI; i18n + typing fixes; Playground ITF plugin; node SVG test script


hand-dot:
Thanks for the contribution. Closing this PR as part of backlog cleanup, without merging.

It combines several independent barcode changes (QR/PDF417 controls, ITF support, SVG rendering, and UI changes), along with generated PDF samples. We would prefer to review these as smaller, focused PRs against current main, with regression coverage for the relevant Node and browser paths.

This does not mean the individual feature requests are rejected. This PR remains available as a reference; please open focused follow-ups for the features you still need.

PR #1188, opened 2025-09-29
fix: Fix calculate target page index in function page breaking.

HoangKien02:


hand-dot:
On it

hand-dot:
Released as 5.4.6

Issue #1220, opened 2025-10-08
template2SchemasList() has an issue with position calculation


hank7444:
### Describe the bug

<img width="576" height="401" alt="Image" src="https://github.com/user-attachments/assets/0f2adf5d-a5e7-4092-ba38-c8f39c7e7bf5" />

When an element exceeds the page width, it adds its original x position again, effectively pushing the element even further to the right — which can cause it to go more out of bounds.
Similarly, the same issue applies to the yEdge calculation.

It should be:
```
if (xEdge > width) {
  value.position.x -= (xEdge - width);
  if (value.position.x < 0) value.position.x = 0; // 防止負數
}
if (yEdge > height) {
  value.position.y -= (yEdge - height);
  if (value.position.y < 0) value.position.y = 0;
}
```

### To Reproduce

N/A

### Expected behavior

N/A

### Your Environment

```markdown
- pdfme package(@pdfme/generator or @pdfme/ui): @pdfme/ui
- pdfme version: 5.2.16
- Operating system: Mac OS
- Node.js version or Browser name & version: Chrome 134.0.6998.118
```

### Your Error Log

```shell
N/A
```

hand-dot:
released as 5.4.7

PR #1228, opened 2025-11-01
fix(ui): correct element position clamping in template2SchemasList

Romanafanasyev:
Fixes the incorrect position adjustment in `template2SchemasList()` where elements exceeding page boundaries were pushed further out of bounds.

This resolves the bug described in #1220.

Changes:
- Replace `+= diff` with proper clamping using `Math.max(0, pageDim - elementDim)`
- Ensure elements stay within page bounds without going negative

Tested locally with sample templates containing out-of-bounds elements.


hand-dot (approved):
LGTM!

PR #1233, opened 2025-11-18
fix: make line elements draggable when height is very small

### LINKED ISSUE #1006: Line cannot be drag and move.
### Describe the bug

The line only can change it location by setting the X and Y manually, it cannot be drag and move although I use it in the playground also. but it can adjust the width and height

### To Reproduce

direct use it in the playgound is same

### Expected behavior

The line only can change it location by setting the X and Y manually, it cannot be drag and move although I use it in the playground also. but it can adjust the width and height

### Your Environment

```markdown
- pdfme version:
- Operating system:
```

### Your Error Log

```shell
No error log for this bug, you can try in the playground
```

### Additional context

_No response_

EedgeY:
## Description
This PR significantly improves the usability of the line schema in the UI by making thin lines easier to select and drag.

## The Problem
Previously, lines with a small height (e.g., 0.5) were extremely difficult to grab and move with the mouse, as their clickable target area was too small. This resulted in a frustrating user experience, especially for users who need precise layouts.

## The Solution
A transparent "hit area" has been introduced over the visible line element. This invisible area is larger than the line itself, providing a generous target for mouse interactions without altering the visual appearance of the line.

- Adds a dedicated `hitArea` element for mouse events.
- The visible line element now ignores pointer events (`pointer-events: none`).
- This change **only affects the UI rendering**; the PDF output remains completely unchanged.

## Demo
Here are videos comparing the behavior before and after the change:

**Before:**

https://github.com/user-attachments/assets/e4cf7aa8-0b20-444f-9999-2b07458ece7a

**After:**

https://github.com/user-attachments/assets/051a852f-6f4c-4d93-bd51-b03e75534977

Fixes #1006

### DIFF
--- a/packages/schemas/src/shapes/line.ts
+++ b/packages/schemas/src/shapes/line.ts
@@ -9,6 +9,7 @@ import { HEX_COLOR_PATTERN } from '../constants.js';
 import { Minus } from 'lucide';
 
 const DEFAULT_LINE_COLOR = '#000000';
+const HIT_POINT_HEIGHT = 16;
 
 interface LineSchema extends Schema {
   color: string;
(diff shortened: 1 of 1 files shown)


hand-dot:
onit

hand-dot (approved):
Coooool

hand-dot:
released as 5.4.8

Issue #706, opened 2025-02-13
Table Page Break Creates Unnecessary Extra Page


lowbits:
### Describe the bug

When using a dynamic table positioned at the bottom of a page, if the table content spans across two pages and there is additional content after the table, the PDF generates an unnecessary blank page. The table items are not efficiently distributed across pages, leading to poor space utilization.

<img width="1644" alt="Image" src="https://github.com/user-attachments/assets/a38c9dcd-43dc-4660-8420-444169ce8c29" />

<img width="441" alt="Image" src="https://github.com/user-attachments/assets/53b97da7-0d9f-4234-a54c-4eeaa00085c4" />

### To Reproduce

1. Create a PDF template with the following structure:
   - Page 1: Content followed by a dynamic table at the bottom
   - Page 2: Additional text content after the table
2. Generate PDF with table data that requires page break (e.g., 3 table rows)

### Expected behavior

- The table should break across pages efficiently without creating an extra page
- Expected layout:
  - Page 1: Content + first table row
  - Page 2: Remaining table rows + the following text content

### Your Environment

```markdown
- pdfme package @pdfme/ui:
- pdfme version: [5.3.5]
- Operating system:
- Node.js v23.6.0
```

### Your Error Log

```shell
NO ERROR
```

baliya367:
I may have found a solution to your problem. You can look here [click](https://javierlopez.eu/captha/) Hope this helped

gosangam:
Is there any progress on this?
I'm also facing this same issue

Issue #629, opened 2024-11-12
Add repeatHead option for table with dynamic data


moraes-leonardo:
### Is your feature request related to a problem? Please describe.

Because I've created a staticSchema that goes with the table when another page is generated, the lack of head on table repetitions leaves it a little confusing.

### Describe the solution you'd like

I would like to be able to specify a parameter on my table attribute, to make it repeat the head on each generated page.

### Describe alternatives you've considered

A boolean parameter that defines if the head will be repeated or not

adelattre49:
Any news about this feature ?

Do you have found a workaround ?

Yuhi-Sato:
Perhaps I am facing a similar problem.

### Is your feature request related to a problem? Please describe.
In some cases, such as quotations and other forms functions, it is important to have headers that display columns even in a paginated table.

For example, if a table column displays “Amount,” which is the result of multiplying a “Unit Price” by a quantity, it may be confusing if the paginated table does not have a column header. Also, some companies may require this as a layout requirement.

### Describe the solution you'd like
As @moraes-leonardo suggested, I suggest adding a repeatHead property, which defaults to a boolean of false, to a schema whose type is table.

For example

```template.json
{
  "schemas": [
    [
      {
        "name": "hoge",
        "type": "table",
        ...
        "showHead": true,
        "repeatHead": true
        ...
      }
    ]
  ]
}
```

### Describe alternatives you've considered


### Additional context
Example of `TableSchema` interface changes at `packages/schemas/src/tables/types.ts`

```diff
export interface TableSchema extends Schema {
  showHead: boolean;
+ repeatHead: boolean;
  head: string[];
  headWidthPercentages: number[];

  tableStyles: {
    borderColor: string;
    borderWidth: number;
  };
  headStyles: CellStyle;
  bodyStyles: CellStyle & { alternateBackgroundColor: string };
  columnStyles: {
    alignment?: { [colIndex: number]: ALIGNMENT };
  };
}
```

jonelofficial:
Any updates on the implementation of this feature? I'm following along and would love to know if there's any progress

PR #1238, opened 2025-11-19
feat(table): add headerRepeat option and fix dynamic layout

EedgeY:
### Overview

- Improve `dynamicTemplate` page-breaking logic so that, when there are many pages, table rows no longer overlap with the bottom margin or subsequent elements.
- Add a `headerRepeat` option to dynamic tables so that header rows can be repeated on each generated page when a table spans multiple pages.

### Related Issues

- Closes [#706 – Table Page Break Creates Unnecessary Extra Page](https://github.com/pdfme/pdfme/issues/706)
- Closes [#629 – Add repeatHead option for table with dynamic data](https://github.com/pdfme/pdfme/issues/629)

### Background

- Previously, when a dynamic table was placed near the bottom of a page and spanned to the next page along with additional content, the layout logic could:
  - Let table rows intrude into the bottom margin.
  - Push following content inefficiently, sometimes resulting in an unnecessary extra blank page (as described in [#706](https://github.com/pdfme/pdfme/issues/706)).
- In addition, there was no way to repeat the table header on every page for long tables, which made multi-page tables harder to read (requested in [#629](https://github.com/pdfme/pdfme/issues/629)).
- This PR introduces:
  - A “long page → page-splitting” layout flow that considers all pages’ schemas at once.
  - A `headerRepeat` option and header-aware height calculation so headers can be repeated cleanly when tables are split across pages.

### Changes

- **`@pdfme/common`**
  - Reworked `getDynamicTemplate`:
    - Flatten all template pages’ schemas into a single “long page” with `flattenSchemasWithOffset`, preserving vertical order but offsetting `position.y` across pages.
    - Build a `LayoutNode` tree for that long page, then call `breakIntoPages` once to split it into multiple physical pages.
    - Reconstruct a new `Template` via `createNewTemplate`, grouping split segments of the same `schema.name` into unified schemas per page, with:
      - `schema.__bodyRange` to mark which body rows (for tables) belong to the current page.
      - `schema.__isSplit` to indicate if the table is continued from a previous page.
  - Enhanced `breakIntoPages`:
    - Track `currentPageIndex` to avoid moving elements back to earlier pages.
    - When an element does not fit the remaining space, move it forward page-by-page, handling the case where an element’s own height is larger than a single page’s content height.
    - Ensures elements (including table parts and following text) don’t overlap and are compactly distributed across pages.
  - Tests in `dynamicTemplate.test.ts`:
    - Adjusted existing multi-page expectations to reflect the new, more precise splitting behavior.
    - Added `"Long page flow (cross-template-page)"` test to verify:
      - A table starting at the bottom of page 1 and text on page 2 are laid out across two pages without generating an extra blank page.
      - The table’s continuation and text appear together on the second page in the correct order.

- **`@pdfme/schemas`**
  - **Types and schema behavior**
    - Extended `TableSchema` with an optional `headerRepeat?: boolean`.
    - In `createSingleTable`:
      - Preserve alternating background logic when `__bodyRange.start` is odd.
      - Compute `schema.showHead` so that:
        - If `schema.showHead === false`, headers are never shown.
        - Otherwise, show the header when:
          - The table is not split (`!schema.__isSplit`), or
          - The table is split but `headerRepeat === true` (i.e., explicitly requested).
  - **Dynamic heights for tables**
    - `getDynamicHeightsForTable` now:
      - Builds `baseHeights`:
        - If `showHead` is true: heights of all rows (head + body).
        - If `showHead` is false: uses `[0, ...bodyRowHeights]` to align body rows without a rendered head.
      - Computes `headerHeight` using `table.getHeadHeight()` when `showHead` is true.
      - Enables header repetition only when:
        - `schema.headerRepeat` is true,
        - `basePdf` is a blank PDF (`isBlankPdf(basePdf)`), and
        - `headerHeight > 0`.
      - Simulates row-by-row placement across pages:
        - Tracks the current page index, current Y position, and the number of rows placed on the current page.
        - For each body row, if it is the first row on a new page (after the initial one), adds `headerHeight` to that row’s effective height so the header’s vertical space is accounted for.
        - Uses a small safety margin to avoid off-by-a-fraction overlap at page boundaries.
        - When a row (including header + row height) doesn’t fit:
          - If we are at the top of a page and the row-plus-header is still too tall, place it on that page anyway.
          - Otherwise, move to the next page and retry.
      - Returns the adjusted heights array, which guarantees that:
        - Headers are counted where they will be rendered.
        - Rows do not overlap the bottom margin or spill inefficiently into extra blank pages.
  - **Prop panel and defaults**
    - `propPanel`:
      - Added `headerRepeat` boolean field next to `showHead` (splitting the span into 12 + 12).
    - `defaultSchema`:
      - Set `headerRepeat: false` by default to keep existing behavior unchanged unless explicitly enabled.

- **`@pdfme/common` – schema validation**
  - Extended `Dict` in `schema.ts` to include `'schemas.table.headerRepeat'` so that the new i18n key is validated.

- **`@pdfme/ui` – i18n**
  - Added translations for `'schemas.table.headerRepeat'` in all supported languages (en, zh, ja, ko, ar, th, it, pl, de, es, fr) so the prop panel label appears correctly in the Designer UI.

### Impact

- **User-facing behavior**
  - Multi-page dynamic tables:
    - No longer overlap the bottom margin or leave unnecessary blank pages when combined with subsequent content.
    - Can now repeat headers on each page when `headerRepeat` is enabled, improving readability on long tables.
  - The default `headerRepeat` is `false`, so existing templates keep their visual behavior unless the new option is explicitly turned on.
- **Compatibility**
  - `TableSchema` is extended in a backward-compatible way (optional field with a default).
  - Header repetition is only enabled for blank `basePdf` layouts; behavior with custom PDFs remains consistent with previous limitations.

###Snapshot

before
before
[before.pdf](https://github.com/user-attachments/files/23630096/before.pdf)

after
after

[after.pdf](https://github.com/user-attachments/files/23630092/after.pdf)
[before.pdf](https://github.com/user-attachments/files/23630093/before.pdf)

### Sample
[sample-template.json](https://github.com/user-attachments/files/23629857/sample.json)

[sample-data.json](https://github.com/user-attachments/files/23629863/sample-data.json)


hand-dot:
on it

hand-dot (approved):
LGTM! 

I performed manual tests and performance tests just in case, but I would like to merge after testing again tomorrow.

This is the main feature of the next minor version upgrade!
Soooo Good!! Thanks @EedgeY

hand-dot:
Released as 5.5.0!!

PR #1242, opened 2025-11-23
fix(schemas:text): align text background rotation with text

EedgeY:
- **Bug**: When the text was rotated, the position of the text and its background rectangle did not align correctly.  
- **Cause**: The text was rotated around a `pivotPoint` using `rotatePoint`, while the background rectangle relied only on the `rotate` option of `drawRectangle`, which rotates around the bottom-left corner.  
- **Fix**: Make the background rectangle use the same rotation logic as the text by rotating its position around the `pivotPoint` with `rotatePoint` before drawing it.

Before
[before.pdf](https://github.com/user-attachments/files/23694130/31f330cc-e0b8-4b12-8e49-dd9aad9e2a85.pdf)

After
[after.pdf](https://github.com/user-attachments/files/23694133/128470a5-14ac-4a36-8fbc-19148ff7446f.pdf)

### DIFF
--- a/packages/schemas/src/text/pdfRender.ts
+++ b/packages/schemas/src/text/pdfRender.ts
@@ -119,9 +119,17 @@ export const pdfRender = async (arg: PDFRenderProps<TextSchema>) => {
     opacity,
   } = convertForPdfLayoutProps({ schema, pageHeight, applyRotateTranslate: false });
 
+  const pivotPoint = { x: x + width / 2, y: pageHeight - mm2pt(schema.position.y) - height / 2 };
+
   if (schema.backgroundColor) {
     const color = hex2PrintingColor(schema.backgroundColor, colorType);
-    page.drawRectangle({ x, y, width, height, rotate, color });
+    if (rotate.angle !== 0) {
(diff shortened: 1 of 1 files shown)


EedgeY:
I found that the same rotation/background misalignment bug exists in other schemas as well.
I opened the PR, but please hold off on reviewing it for now. I’m investigating how to unify the handling to fix this across all relevant schemas.

EedgeY:
My findings confirm that the bug fix implemented in this PR for the text schema effectively resolves the issues present in all other related schemas as well.

Therefore, I agree that proceeding with the review is the appropriate next step. Please proceed at your convenience.

before
[787794c2-4d45-4d0d-aac1-d759db01009c.pdf](https://github.com/user-attachments/files/23695453/787794c2-4d45-4d0d-aac1-d759db01009c.pdf)

after
[0ee90b05-3347-44ca-983c-9a53623a9949.pdf](https://github.com/user-attachments/files/23695461/0ee90b05-3347-44ca-983c-9a53623a9949.pdf)

hand-dot (approved):
LGTM!

hand-dot:
released as 5.5.2

PR #1249, opened 2025-11-28
fix: resize window down scales canvas height too

### LINKED ISSUE #1248: Resizing DOWN should reduce the overall page canvas size
### Describe the bug

When you resize the screen up it scales the content up to fit, using more visual real-estate (which is great!).

However, when you scale down, the overall height of the canvas doesn't change, which means the left and right scrollbars becomes scrollable, even thought their contents is already scrollable.

Ideally, the canvas should reduce in size so that the sidebars fit the page as before. I don't think we should zoom the PDF back out though, I think it should probably remain the same and the user can zoom out 🤔 
We can try it and see. 

https://github.com/user-attachments/assets/ad3527dd-b633-449f-8210-20165b1b5a8b

### To Reproduce

Start with a small window
Resize to make it much larger
Resize to make it much smaller

### Expected behavior

The sidebars should not have a double-scrollbar introduced by resizing down.

### Your Environment

```markdown
- pdfme version: latest main
- Operating system: Mac 
- Node.js version or Browser name & version: Chrome
```

### Your Error Log

```shell
-
```

### Additional context

_No response_

peteward:
fixes #1248 

BEFORE - scale up increases size, but scale down doesn't.
This is because element.clientHeight does not return the visible height, but the overall height of the element - even if it's not visible because it's being scrolled. Therefore the height (and width) can never decrease.

https://github.com/user-attachments/assets/8bc59c67-f4cc-4414-981c-8625f2f39f1e

AFTER - scaling up and down consistently on resize using boundingRect

https://github.com/user-attachments/assets/72ae9425-8516-461f-b930-cdfd1c910338

### DIFF
--- a/packages/ui/src/class.ts
+++ b/packages/ui/src/class.ts
@@ -40,10 +40,19 @@ export abstract class BaseUIClass {
     if (!this.domContainer) {
       return;
     }
+
+    const rect = this.domContainer.getBoundingClientRect();
+    const vw = window.innerWidth;
+    const vh = window.innerHeight;
+
+    const visibleWidth = Math.max(0, Math.min(rect.right, vw) - Math.max(rect.left, 0));
+    const visibleHeight = Math.max(0, Math.min(rect.bottom, vh) - Math.max(rect.top, 0));
+
     this.size = {
-      height: this.domContainer.clientHeight || window.innerHeight,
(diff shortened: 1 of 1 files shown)


hand-dot:
Released as 5.5.5

PR #1250, opened 2025-11-28
feat: default text schema line height to 1.3

### LINKED ISSUE #1017: Default line-height to something more suitable

The default pdfme line-height is 1.0, which is usually not the best for any content that is multi-line.

I recommended we set a default line-height of between 1.2 and 1.5 (people on the web seem to offer different advice here).

Whilst it obviously depends on the font what a good value is, I think larger than 1.0 would be better.

Any thoughts what it should be?

-

### Additional context

_No response_

peteward:
Closes #1017 

Changing the line-height affected most of the existing test templates where it wasn't explicitly specified - although only very minor vertical differences. I've made adjustments to compensate in some templates to ensure they still look good, then regenerated the snapshots.


hand-dot:
Hi @peteward !
is this urgent? I'm debating whether this fix should be included in a major version update, as not setting lihe-height will result in a slight visual difference compared to previously created PDFs.

peteward:
it's not urgent, no  :)

PR #1288, opened 2025-12-04
chore: replace npm-run-all with npm-run-all2

ryoppippi:
npm-run-all is unmaintained anymore and there are some critical bugs.
npm-run-all2 is a fork of npm-run-all and it is actively maintained.

https://www.npmjs.com/package/npm-run-all2

p.s. i think it is better to switch our package manager to pnpm because we can reduce lots of dependencies including npm-run-all2, but it depends on what you want!!

### DIFF
--- a/package.json
+++ b/package.json
@@ -32,7 +32,7 @@
   ],
   "scripts": {
     "postinstall": "./scripts/link-workspaces.sh",
-    "clean": "npm-run-all2 --parallel clean:*",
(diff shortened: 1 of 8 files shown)


hand-dot (approved):
LGTM!

hand-dot:
Released as 5.5.7 🚀

ryoppippi:
oh i forgot to updae the lockfile! thanks

Issue #1183, opened 2025-09-09
Image orientation issue (EXIF rotation not respected)


sajjadramezani1373:
### Describe the bug

When inserting an image into a PDF using pdfme, images with an EXIF Orientation tag are not displayed correctly.

For example, I have an image with Orientation: Rotate 180.
- When I open this image in Windows/Mac viewer or other software, it is displayed correctly (because most viewers respect EXIF orientation).
- But when I render it with pdfme, the image is rotated 180° inside the PDF.

### To Reproduce

**Attached files:**

![Image](https://github.com/user-attachments/assets/adb34ec9-0e14-42f5-8d19-4f3c2ca6a33f)
<img width="1888" height="784" alt="Image" src="https://github.com/user-attachments/assets/45bc1977-06c7-4389-a0d4-ebfb0bd6f2b9" />
<img width="1899" height="906" alt="Image" src="https://github.com/user-attachments/assets/086369c9-748e-496e-8c77-6cda2cdb78d6" />

### Expected behavior

The image should be displayed with the correct orientation (like in OS viewers, browsers, etc.), respecting EXIF metadata.

### Your Environment

```markdown
**Environment:**

node.js version: 20.17.0
"@pdfme/common": "^5.3.15",
"@pdfme/generator": "^5.3.15",
"@pdfme/schemas": "^5.3.15",
"@pdfme/ui": "^5.3.15",
"next": "^15.2.5",
"react": "^18.3.1",
"react-dom": "^18.3.1",
```

### Your Error Log

```shell
Without error
```

superiority-claw-agent:
Partially addressed in [PDFweave PR #5](https://github.com/IDNTEQ/pdfweave/pull/5) — commit [`5d4d15fa`](https://github.com/IDNTEQ/pdfweave/commit/5d4d15fa).

Shipped: an inline EXIF Orientation parser (~80 LOC, no dependencies) that reads the Orientation tag (1-8) from JPEG APP1 markers. Comprehensive test coverage.

**Not yet shipped:** wiring the parsed orientation into image draw geometry. The schema's `width`/`height` need to refer to the un-rotated bytes for pdf-lib's `drawImage` to scale correctly, and threading a second rotation through the existing center-pivot correction in `convertForPdfLayoutProps` without regressing existing rotated-image cases needs more careful tuning than the batch budget allowed.

Tracking the wire-up for v0.1.1. Context: pdfme/pdfme#1468.

[...]

PR #1290, opened 2025-12-09
[@pdme/converter] Change Worker loading in local environment to ESM

tetty0217:
## Overview
While developing a feature that imports and uses `converter` locally, I noticed an error during module resolution when initializing the Worker.

### Before Fix
This is an error log when executing `pdf2img` in a browser runtime, for example.
```
Uncaught SyntaxError: The requested module '/node_modules/.vite/deps/pdfjs-dist_build_pdf__worker__entry__js.js?v=0b0a5bfb' does not provide an export named 'default' (at index.browser.ts:3:8)Understand this error
```

### After Fix
Executing `pdf2img` in a browser runtime now correctly resolves source modules without errors.
**pdf2img conversion in Reference is successful✨**

### Reference
This is reference code that was incorporated into the Playground for confirmation, although it is not an actual feature.

```tsx
const Converter = () => {
  const [url, setUrl] = useState('');

  const uiRef = useRef<HTMLDivElement | null>(null);

  const navItems: NavItem[] = [
    {
      label: '',
      content: React.createElement(ExternalButton, {
        href: 'https://github.com/pdfme/pdfme/issues/new?template=template_feedback.yml&title=TEMPLATE_NAME',
        title: 'Feedback this template',
      }),
    },
  ];

  useEffect(() => {
    const init = async () => {
      const template = templateJson as Template;

      const pdf = await generate({
        template: template,
        inputs: getInputFromTemplate(template),
        options: getFontsData(),
        plugins: getPlugins(),
      });

      if (pdf) console.log('✨ generate pdf');

      const image = await pdf2img(pdf.buffer);
      if (image) console.log('✨ convert image');

      const blob = new Blob(image, { type: 'image/png' });
      const url = URL.createObjectURL(blob);

      setUrl(url);
    };

    init();
  }, []);

  return (
    <>
      <NavBar items={navItems} />
      <button
        style={{ width: '300px', height: '100px' }}
        onClick={() => {
          const confirmed = window.confirm('OK?');
          if (confirmed) {
            window.open(url);
            window.setTimeout(() => {
              URL.revokeObjectURL(url);
            }, 100);
          }
        }}
      >
        Go!
      </button>
      <div ref={uiRef} className="flex-1 w-full" />
    </>
  );
};

```

## Test
- [x] The `converter` functionality works on the Playground's dev server.
- [x] The `converter` functionality continues to work without changes in the Playground's preview.
- [x]  

## Note
- The fix could not be implemented because the browser module in the converter was being built for CJS. Therefore, we confirmed that https://github.com/pdfme/pdfme/pull/1122 is planned and modified the tsconfig.

## Triage disposition (2026-09-13)

Closed as superseded by #1558, which replaced the pdfjs-dist renderer with clawpdf/PDFium. Current main uses a dedicated clawpdf module worker, so the pdfjs worker import addressed by this PR is no longer present. This does not mark the separate EXIF-orientation issue #1183 or the clawpdf/Vite integration issue #1560 as resolved.

### DIFF
--- a/packages/converter/src/index.browser.ts
+++ b/packages/converter/src/index.browser.ts
@@ -1,10 +1,11 @@
 import * as pdfjsLib from 'pdfjs-dist';
-// @ts-expect-error - PDFJSWorker import is not properly typed but required for functionality
-import PDFJSWorker from 'pdfjs-dist/build/pdf.worker.entry.js';
 import { pdf2img as _pdf2img, Pdf2ImgOptions } from './pdf2img.js';
 import { pdf2size as _pdf2size, Pdf2SizeOptions } from './pdf2size.js';
 
-pdfjsLib.GlobalWorkerOptions.workerSrc = PDFJSWorker as unknown as string;
+pdfjsLib.GlobalWorkerOptions.workerSrc = new URL(
+  'pdfjs-dist/build/pdf.worker.entry.js',
+  import.meta.url,
+).toString();
 
 function dataURLToArrayBuffer(dataURL: string): ArrayBuffer {
   // Split out the actual base64 string from the data URL scheme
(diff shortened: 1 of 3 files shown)


tetty0217:
@hand-dot 
Hi😃
There is no urgency for this correction. Please take a look when you have a moment!

As a follow-up to this PR, we plan to address the resolution of the next issue（ https://github.com/pdfme/pdfme/issues/1183 ）

Issue #1291, opened 2025-12-10
Incorrect Object Positions on Pages other than the First Page


mmohr-alta4:
### Describe the bug

When creating a multi-page Document in the Designer and generating a PDF from it, the positions on all pages after the first one seem to be shifted.
[document.pdf](https://github.com/user-attachments/files/24077543/document.pdf)
PDF:
<img width="479" height="839" alt="Image" src="https://github.com/user-attachments/assets/b100198f-6fd9-4413-8669-fb1b605b5aa9" />

Designer (correct):

<img width="334" height="603" alt="Image" src="https://github.com/user-attachments/assets/ac837bec-8678-4ec4-ad60-3d6a90780801" />

### To Reproduce

Use this template.
Generate a PDF
`
{"schemas":[[{"name":"field1","type":"ellipse","position":{"x":19.54,"y":65},"width":62.5,"height":37.5,"rotate":0,"opacity":1,"borderWidth":1,"borderColor":"#000000","color":"","readOnly":true,"radius":0,"required":false,"content":""}],[{"name":"field2","type":"ellipse","position":{"x":20.85,"y":65},"width":62.5,"height":37.5,"rotate":0,"opacity":1,"borderWidth":1,"borderColor":"#000000","color":"","readOnly":true,"radius":0,"required":false}]],"basePdf":{"width":210,"height":297,"padding":[20,10,20,10],"staticSchema":[{"name":"field1","type":"rectangle","position":{"x":21.96,"y":26.99},"width":62.5,"height":37.5,"rotate":0,"opacity":1,"borderWidth":1,"borderColor":"#000000","color":"","readOnly":true,"radius":0,"required":false}]},"pdfmeVersion":"5.5.6"}
`

### Expected behavior

The positions should be correct on all pages of the PDF

### Your Environment

```markdown
- pdfme package(@pdfme/generator or @pdfme/ui): Playground
- pdfme version: 5.7.0
```

### Your Error Log

```shell
No error log
```

hand-dot:
Fixed in 5.5.8. Can you check it? If it's not fixed, please reopen it.

Issue #1292, opened 2025-12-10
Inconsistent page height calculation in flattenSchemasWithOffset


Thorben91:
### Describe the bug

In `@pdfme/common/src/dynamicTemplate.ts`, `flattenSchemasWithOffset` uses variable page heights, but `breakIntoPages.calculateNewY` expects constant page heights.

`flattenSchemasWithOffset` (line 306):

```typescript
const getPageHeight = (pageIndex: number) => basePdf.height - paddingBottom - (pageIndex > 0 ? paddingTop : 0);
```

`breakIntoPages.calculateNewY` (line 172):

```typescript
const newY = y - pageIndex * (basePdf.height - paddingTop - paddingBottom);
```

Result: Schemas on page 1+ are shifted down by `paddingTop`.

### To Reproduce

1. Create a multi-page PDF template with `paddingTop > 0`
2. Add schemas that span multiple pages
3. Generate the PDF
4. Observe that schemas on page 1+ are shifted down by `paddingTop` pixels

### Expected behavior

Schemas should maintain their Y-position across all pages.

### Your Environment

```markdown
- pdfme package: @pdfme/common
- pdfme version: 5.5.7
- Operating system: Windows 10
- Node.js version or Browser name & version: 22.17.0
```

### Your Error Log

```shell
No error log — this is a silent calculation bug that causes incorrect positioning.
```

### Additional context

This bug was identified with the assistance of Claude Opus 4.5. I have created and applied a local patch using `patch-package` which fixes the issue by modifying `getPageHeight` in `flattenSchemasWithOffset` to use consistent page heights:

```diff
// In flattenSchemasWithOffset (line ~306)
- const getPageHeight = (pageIndex: number) => basePdf.height - paddingBottom - (pageIndex > 0 ? paddingTop : 0);
+ const getPageHeight = () => basePdf.height - paddingTop - paddingBottom;
```

hand-dot:
Fixed in 5.5.8. Can you check it? If it's not fixed, please reopen it.

PR #1294, opened 2025-12-12
fix(dynamicTemplate): rewrite layout engine for accurate multi-page table rendering

EedgeY:
## Fixes

issues 
 #1291
#1292

## Overview

This PR addresses a critical layout issue where elements on the 2nd page (and beyond) of a multi-page template were incorrectly shifted upwards due to miscalculated offsets. We have simplified the dynamicTemplate calculation logic to fix this behavior and ensured that dynamic table row heights are recalculated immediately upon form input confirmation in the UI.

## Key Changes
1. Layout & Offset Logic Refactoring (Multi-page Fix)
Fix 2nd+ Page Y-Coordinates: Fixed a bug where elements created in the Designer on subsequent pages were being "pushed up" (incorrect Y-coordinates).
Simplified Calculation Algorithm: Refactored the calculation method to use a simpler, sequential offset-based approach. The totalYOffset is now accurately tracked across pages to ensure correct positioning.
Floating Point Tolerance: Added EPSILON to prevent precision errors during page break calculations.
2. Table & Data Handling
Prevent Orphaned Headers: Adjusted page break logic so table headers move to the next page with the first data row if there isn't enough space, preventing headers from being left alone at the bottom of a page.
Fix __bodyRange Indexing: Corrected indexing to properly map dynamicHeights to table body data.
3. UI/Preview Updates
Immediate Recalculation: Updated the preview logic to trigger height calculations immediately when form inputs are confirmed. This ensures the table row heights in the preview match the actual content dynamically.
Logic Flow
The new layout generation process is structured as follows:

### Technical Details
Implemented sequential offset-based layout algorithm.
Updated tests to reflect the corrected multi-page layout behavior.

[sample-template.json](https://github.com/user-attachments/files/24123201/sample-template.json)
[sample-data500.json](https://github.com/user-attachments/files/24123116/sample-data500.json)
[sample-data1000.json](https://github.com/user-attachments/files/24123117/sample-data1000.json)
[issues-1291-template.json](https://github.com/user-attachments/files/24123120/issues-1291-template.json)

### DIFF
--- a/packages/common/src/dynamicTemplate.ts
+++ b/packages/common/src/dynamicTemplate.ts
@@ -284,22 +284,9 @@ export const getDynamicTemplate = async (
       }
     }
 
-    // Check if this page has any height changes
-    let pageHasChange = false;
-    for (const item of items) {
-      const totalHeight = item.dynamicHeights.reduce((a, b) => a + b, 0);
-      if (Math.abs(totalHeight - item.height) > EPSILON) {
-        pageHasChange = true;
-        break;
-      }
-    }
-
-    if (pageHasChange) {
-      const processedPages = processDynamicPage(items, orderMap, contentHeight, paddingTop);
-      resultPages.push(...processedPages);
-    } else {
-      resultPages.push(items.map((item) => item.schema));
-    }
+    // Process all pages independently (no cross-page offset propagation)
+    const processedPages = processDynamicPage(items, orderMap, contentHeight, paddingTop);
+    resultPages.push(...processedPages);
   }
 
   removeTrailingEmptyPages(resultPages);


hand-dot:
onit

EedgeY:
### Changes

- Process template pages independently instead of flattening all schemas into a continuous coordinate system
- Static pages (no height changes) are added as-is after dynamic pages, without offset propagation

### Behavior

- Pages with dynamic content (e.g., tables): Full layout calculation with page breaks
- Pages without height changes: Copied directly without layout recalculation

This reduces computation cost by skipping detailed layout calculations for pages that don't need them.

hand-dot:
Released as [5.5.8](https://github.com/pdfme/pdfme/releases/tag/5.5.8)

PR #1295, opened 2025-12-14
add dependabot automerge for minor + patch updates

peteward:
I've set this up for minor and patch updates, which should be safe (if we don't auto-release). But you might prefer just patch.


hand-dot (approved):

hand-dot:
Thank you @peteward !

PR #1304, opened 2026-01-03
fix(schemas): preserve font weight in Designer read-only text

sohailxdev:
### Problem
In the Designer, text rendered in read-only mode loses its font weight
after blur because characters are rendered as individual `<span>` elements
without explicitly setting the resolved font-family.

### Solution
Apply the resolved font-family (using postscriptName when available, otherwise schema.fontName)
 to each character span in read-only mode to ensure
consistent font rendering.

### Impact
- Fixes bold / font-weight mismatch in Designer
- No change to PDF output
- No breaking changes
- Scoped only to UI rendering

### DIFF
--- a/packages/schemas/src/text/uiRender.ts
+++ b/packages/schemas/src/text/uiRender.ts
@@ -81,7 +81,7 @@ export const uiRender = async (arg: UIRenderProps<TextSchema>) => {
   );
 
   const processedText = replaceUnsupportedChars(value, fontKitFont);
-  const resolvedFontName = fontKitFont.postscriptName || schema.familyName;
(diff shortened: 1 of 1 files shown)


sohailxdev:
Hi maintainers 👋  
This PR includes a small UI-only fix in the Designer text renderer.  
The failing Vercel check appears to be due to forked PR workflow restrictions.  
Happy to make any changes if needed — thanks!

sohailxdev:
Good catch 👍  
You're right — `schema.familyName` does not exist on `TextSchema`.

I've updated the fallback logic to correctly use:
`fontKitFont.postscriptName || schema.fontName`

This ensures the font-family is always resolved correctly in read-only Designer mode.
Thanks for pointing it out!

Issue #443, opened 2024-03-11
Move form-render dependency from common package to ui package


aaronmcdavid:
### Is your feature request related to a problem? Please describe.

Alibab's form-render package has a set of dependencies that are both not current and not relevant to the generation of PDFs.  To prevent the unnecessary incorporation of old, unused packages, this dependency can be moved closer to where it is needed.

### Describe the solution you'd like

Move the form-render dependency out of the common package to the ui package.  This also appears to entail a move of the types PropPanelSchema and PropPanelWidgetProps to the ui package.  These seem to have no bearing on the common package.

### Describe alternatives you've considered

I've also submitted a request to alibaba for them to update form-render, but it still includes a React color picker component, which doesn't seem necessary to generate PDFs.

hand-dot:
ref: https://github.com/pdfme/pdfme/issues/410

hand-dot:
Hey @aaronmcdavid !
Thank you for the suggestion.
There have been reports of a [similar issue](https://github.com/pdfme/pdfme/issues/410) in the past.
I want to address this issue. 

But the reason form-render is in the common package is because the schema package needs form-render's types for the definition of plugins it uses.
And the schema package is used in the generator package.

I don't think our current approach is the best.
As you said, there are dependencies that are not relevant to PDF generation, which is not good.

However, if we move it to the ui package, we need to think about how to use form-render's types in the schema package.

The most ideal solution would be for form-render to use the latest @rc-component/color-picker.
Could we possibly wait for a response to your issue for a while?

---

ref: https://github.com/alibaba/x-render/issues/1496

hand-dot:
@sebimarkgraf 
>  Did you consider vendoring the types from antd that are really needed in the common package?

How exactly should we do this? I'd like to try it, but I don't know how to go about it. Could you please explain?

thiagotognoli:
> @sebimarkgraf
> 
> > Did you consider vendoring the types from antd that are really needed in the common package?
> 
> How exactly should we do this? I'd like to try it, but I don't know how to go about it. Could you please explain?


Vendoring the necessary types from Ant Design (antd) into a common package can be a strategic approach, especially if you're dealing with a large codebase or need to manage dependencies more effectively. Here are some considerations:

Pros of Vendoring Types
Control Over Types: You can customize or extend types as needed, ensuring they align perfectly with your application.
[...]

PR #1307, opened 2026-01-06
fix(common): mark antd and form-render as optional peer dependencies in common

felhe:
As suggested in https://github.com/pdfme/pdfme/issues/443, this pull request makes a small change to the `packages/common/package.json` file, marking `antd` and `form-render` as optional peer dependencies.

From https://github.com/pdfme/pdfme/issues/443#issuecomment-3427184466:

> The interfaces from antd and form-render have already been marked as optional in a previously merged PR (https://github.com/pdfme/pdfme/pull/477), so common is functional even if these packages aren’t installed.
> 
> However, the main issue remains: although both dependencies have already been moved from dependencies to peerDependencies, they are still automatically installed alongside the common package. This is because from npm v7 onwards, all peer dependencies are installed by default (see [docs](https://docs.npmjs.com/cli/v9/configuring-npm/package-json#peerdependencies)).
> 
> Both dependencies can be marked as optional within the common package using peerDependenciesMeta.
> According to https://github.com/npm/rfcs/issues/221, optional peer dependencies are not installed automatically.

Both packages are still included in the ui package.json, so they are still installed during development and when using the ui package, but they are skipped when *only* common is needed (e.g. in NodeJS environments), preventing build issues.

Related issues: https://github.com/pdfme/pdfme/issues/474, https://github.com/pdfme/pdfme/issues/443

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -12766,6 +12766,14 @@
       "peerDependencies": {
         "antd": "^5.11.2",
         "form-render": "^2.2.20"
+      },
+      "peerDependenciesMeta": {
+        "antd": {
+          "optional": true
+        },
+        "form-render": {
+          "optional": true
+        }
       }
     },
     "packages/converter": {
--- a/packages/common/package.json
+++ b/packages/common/package.json
@@ -57,6 +57,14 @@
     "antd": "^5.11.2",
     "form-render": "^2.2.20"
   },
+  "peerDependenciesMeta": {
+    "antd": {
+      "optional": true
+    },
+    "form-render": {
+      "optional": true
+    }
+  },
   "jest": {
     "resolver": "ts-jest-resolver",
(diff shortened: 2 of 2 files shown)


GabrielCousin:
Hey!
Any chance this could be merged?
We have the same issue and this looks like a great trade-off.
Thanks a lot for maintaining and contributing ♥️

hand-dot:
@copilot resolve the merge conflicts in this pull request

hand-dot:
Thanks for the PR. The dependency issue this was addressing has now been resolved on `main` by #1420, which removed the `antd` and `form-render` peer dependencies from `@pdfme/common` entirely.

Because those peer dependencies are no longer present, adding `peerDependenciesMeta` here is no longer needed, and this branch is now outdated/conflicting. Closing this as superseded by #1420.

PR #1349, opened 2026-02-26
fix(pdf-lib): use Node.js native crypto for AES-256 decryption

### LINKED ISSUE #1348: AES256Cipher._decrypt produces garbled output for V=5/R=5 encrypted PDFs in Node.js
## Bug Description

The JavaScript AES-256 block cipher implementation in `AESBaseCipher._decrypt` (`packages/pdf-lib/src/core/crypto.ts`) produces garbled content streams when decrypting V=5/R=5 (AES-256) encrypted PDFs in Node.js environments.

## Steps to Reproduce

1. Take a PDF encrypted with AES-256 (V=5, R=5, `/Filter /Standard`)
2. Load it with `PDFDocument.load(bytes, { password: '' })` (empty user password)
3. Save with `pdfDoc.save()`
4. Open the saved PDF — all text is missing, fonts are garbled

## Expected Behavior

The saved PDF should contain all original text with properly decrypted font programs and content streams.

## Root Cause

The `_decrypt` method in `AESBaseCipher` (line 788) implements AES block decryption in pure JavaScript. This implementation produces incorrect output for AES-256 (14 rounds), causing all content streams and font programs to be garbled after decryption.

## Workaround / Proposed Fix

In the `dist/node/` build (Node.js environments), replace the JavaScript AES block cipher with Node.js native `crypto.createDecipheriv('aes-256-ecb')`, which uses OpenSSL and is battle-tested:

```typescript
// In AES256Cipher class
class AES256Cipher extends AESBaseCipher {
  private _rawCipherKey: Uint8Array;

  constructor(key: Uint8Array) {
    super();
    this._cyclesOfRepetition = 14;
    this._keySize = 224;
    this._rawCipherKey = new Uint8Array(key);
    this._key = this._expandKey(key);
  }

  _expandKey(cipherKey: Uint8Array) {
    this._rawCipherKey = new Uint8Array(cipherKey);
    // ... existing key expansion ...
  }

  // Override for Node.js — use native crypto instead of buggy JS implementation
  _decrypt(input: Uint8Array, _key: Uint8Array) {
    const { createDecipheriv } = require('node:crypto');
    const decipher = createDecipheriv('aes-256-ecb', this._rawCipherKey, null);
    decipher.setAutoPadding(false);
    return new Uint8Array(decipher.update(Buffer.from(input)));
  }
}
```

This could be conditionally applied only in the `dist/node/` build, keeping the JS implementation for browser environments where `node:crypto` isn't available.

## Environment

- @pdfme/pdf-lib: 5.5.8
- Node.js: 23.x / 24.x
- PDFs: AES-256 encrypted (V=5, R=5) with empty user password

## Current Workaround

We're using a `pnpm patch` that applies the fix to all three dist variants (esm/cjs/node). Happy to submit a PR with a proper source-level fix.

Casperjuel:
## Summary

Fixes #1348

The pure JavaScript AES-256 block cipher in `AESBaseCipher._decrypt` produces **garbled content streams** when decrypting V=5/R=5 (AES-256) encrypted PDFs in Node.js environments. All text disappears and fonts are corrupted after a load+save cycle.

## Root Cause

The JS implementation of AES-256 decryption (14 rounds) in `AESBaseCipher._decrypt` produces incorrect output. This was confirmed by comparing its output against:
- Node.js `crypto.createDecipheriv('aes-256-ecb')` (OpenSSL-backed)
- The NIST FIPS 197 Appendix C.3 test vector

The included test `demonstrates the JS fallback produces wrong output for AES-256` proves the bug by calling the base class `_decrypt` directly and showing it does **not** match the expected plaintext.

## Changes

- **`packages/pdf-lib/src/core/crypto.ts`**:
  - Adds a conditional `require('node:crypto')` at module scope (safe for browsers — guarded by `globalThis.process` check)
  - Stores the raw 32-byte cipher key in `AES256Cipher._rawCipherKey`
  - Overrides `_decrypt` in `AES256Cipher` to use `createDecipheriv('aes-256-ecb')` when Node.js crypto is available
  - Falls back to the existing JS implementation in browser environments

- **`packages/pdf-lib/__tests__/core/crypto.spec.ts`** (new):
  - Tests decryption of random blocks against Node.js crypto
  - Tests the NIST FIPS 197 AES-256 test vector
  - Proves the JS fallback produces incorrect output (the bug)

## Browser Impact

None — the fix only activates when `node:crypto` is available. Browser environments continue using the existing JS implementation. A separate investigation may be needed for the browser-side JS implementation, but browsers typically use pdfjs-dist for PDF rendering which has its own (working) AES implementation.

### DIFF
--- a/packages/pdf-lib/__tests__/core/crypto.spec.ts
+++ b/packages/pdf-lib/__tests__/core/crypto.spec.ts
@@ -1,0 +1,78 @@
+import * as crypto from 'node:crypto';
+
+import { AES256Cipher } from '../../src/core/crypto';
+
+describe('AES256Cipher', () => {
+  it('should correctly decrypt a single AES-256-ECB block', () => {
+    const key = crypto.randomBytes(32);
+    const plaintext = crypto.randomBytes(16);
+
+    const cipher = crypto.createCipheriv('aes-256-ecb', key, null);
+    cipher.setAutoPadding(false);
+    const encrypted = new Uint8Array(cipher.update(plaintext));
+
+    const aesCipher = new AES256Cipher(new Uint8Array(key));
+    const decrypted = aesCipher._decrypt(encrypted, aesCipher._key);
+
+    expect(Buffer.from(decrypted)).toEqual(Buffer.from(plaintext));
+  });
+
+  it('should correctly decrypt multiple random blocks', () => {
+    const key = crypto.randomBytes(32);
+
+    for (let i = 0; i < 10; i++) {
+      const plaintext = crypto.randomBytes(16);
+
+      const cipher = crypto.createCipheriv('aes-256-ecb', key, null);
+      cipher.setAutoPadding(false);
+      const encrypted = new Uint8Array(cipher.update(plaintext));
+
+      const aesCipher = new AES256Cipher(new Uint8Array(key));
+      const decrypted = aesCipher._decrypt(encrypted, aesCipher._key);
+
+      expect(Buffer.from(decrypted)).toEqual(Buffer.from(plaintext));
+    }
+  });
+
+  it('should decrypt NIST AES-256 test vector (FIPS 197 C.3)', () => {
+    const key = new Uint8Array([
+      0x00, 0x01, 0x02, 0x03, 0x04, 0x05, 0x06, 0x07, 0x08, 0x09, 0x0a,
+      0x0b, 0x0c, 0x0d, 0x0e, 0x0f, 0x10, 0x11, 0x12, 0x13, 0x14, 0x15,
+      0x16, 0x17, 0x18, 0x19, 0x1a, 0x1b, 0x1c, 0x1d, 0x1e, 0x1f,
+    ]);
+    const plaintext = new Uint8Array([
+      0x00, 0x11, 0x22, 0x33, 0x44, 0x55, 0x66, 0x77, 0x88, 0x99, 0xaa,
+      0xbb, 0xcc, 0xdd, 0xee, 0xff,
+    ]);
+    const expectedCiphertext = new Uint8Array([
+      0x8e, 0xa2, 0xb7, 0xca, 0x51, 0x67, 0x45, 0xbf, 0xea, 0xfc, 0x49,
+      0x90, 0x4b, 0x49, 0x60, 0x89,
+    ]);
+
+    const aesCipher = new AES256Cipher(key);
+    const decrypted = aesCipher._decrypt(expectedCiphertext, aesCipher._key);
+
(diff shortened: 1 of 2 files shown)


hand-dot:
onit

hand-dot:
Thanks for identifying and fixing this real bug — the NIST test vector proof is very helpful.

However, we can't merge this as-is due to the `require('node:crypto')` approach. This will cause issues with bundlers (webpack, Vite, Rollup) that statically analyze `require()` calls, even when wrapped in try/catch. Since the ESM build is the default for browser consumers, this would break or warn in browser bundles.

We're planning to move to **ESM-only in the next major version**, which will allow us to solve this more cleanly using **conditional exports**:

```jsonc
// package.json
{
  "exports": {
    ".": {
      "node": "./dist/node/index.js",   // static import of node:crypto
      "default": "./dist/browser/index.js" // JS fallback
    }
  }
}
```

With this approach, the Node.js entry point can use a standard `import crypto from 'node:crypto'` — no try/catch, no runtime guards needed.

**Plan:**
1. We'll first ship the ESM-only major version with conditional exports
2. Then we'd love to revisit this fix on top of that foundation

We'll keep this PR open for reference. If you'd like to contribute the updated version after the ESM migration lands, that would be great!

hand-dot:
Thanks again for tracking this down. I re-reviewed this against the current `main` state.

The AES-256 decryption issue and the NIST vector coverage are valuable, but I still don't think we should merge this implementation as-is because it introduces a guarded `require('node:crypto')` into the shared ESM/browser-facing pdf-lib source. That remains risky for bundlers that statically analyze `require()` calls.

Let's keep this PR open as the reference for the bug and test vectors. The implementation path I would prefer is either:

- fix the pure JS AES-256 implementation so Node and browser both benefit, or
- after the Node/browser entry split is in place, add a Node-only implementation using a static `node:crypto` import behind conditional exports.

So: useful investigation, but not mergeable in its current shape.

hand-dot:
Follow-up: I split out the test-only part into #1456.

That PR keeps the useful AES-256 coverage from this discussion (NIST FIPS 197 C.3 vector plus a Node.js crypto comparison) without adding `node:crypto` to production pdf-lib code. It also passes against current `main`, so it should be safe to review independently while we leave this implementation PR open for the future Node/browser split or a pure JS fix if needed.

hand-dot:
Following up on the cleanup: the production-code approach here is not something we want to merge as-is, and the reusable test coverage has been split out into #1456.

Closing this PR as superseded by #1456 for the test coverage, with any future production fix to happen separately through a browser-safe implementation path.

PR #1425, opened 2026-04-12
fix(schemas): correct CMYK color conversion for zero RGB channels

mb-weberon:
## Summary

  The `hex2CmykColor` function in `packages/schemas/src/utils.ts` has an incorrect division-by-zero guard that causes any color with
  a zero RGB channel to produce wrong CMYK values.

  ## Bug

  Lines 139-141 check `r === 0` / `g === 0` / `b === 0` to guard against division by zero. This is wrong — the division is `/ (1 -
  k)`, which is only zero when `k === 1` (pure black). Checking individual RGB channels causes the corresponding CMYK component to be
   incorrectly zeroed.

  ## Impact

  Any color containing a zero RGB channel renders incorrectly in CMYK mode:

  | Input | Expected CMYK | Produced CMYK |
  |---|---|---|
  | `#ffff00` (yellow) | 0, 0, 1, 0 | 0, 0, 0, 0 (white / no ink) |
  | `#ff0000` (red)    | 0, 1, 1, 0 | 0, 0, 0, 0 |
  | `#00ff00` (green)  | 1, 0, 1, 0 | 0, 0, 0, 0 |
  | `#ff8000` (orange) | 0, 0.498, 1, 0 | 0, 0.498, 0, 0 |

  ## Fix

  Replace the three `x === 0` checks with `k === 1`.
  
  - fixed using claude -

### DIFF
--- a/packages/schemas/__tests__/utils.test.ts
+++ b/packages/schemas/__tests__/utils.test.ts
@@ -1,5 +1,5 @@
 import { Schema, mm2pt, pt2mm } from '@pdfme/common';
-import { convertForPdfLayoutProps, rotatePoint, hex2RgbColor, createSvgStr } from '../src/utils.js';
+import { convertForPdfLayoutProps, rotatePoint, hex2RgbColor, hex2PrintingColor, createSvgStr } from '../src/utils.js';
 import { SquareCheck, IconNode } from 'lucide';
 
 describe('hex2RgbColor', () => {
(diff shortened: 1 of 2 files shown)


hand-dot:
Thanks for the clear fix and regression cases. I rechecked this against current `main`; the CMYK bug is still present, but this PR's `pdfme-test` is currently red because of an unrelated outdated-base CLI manifest version check.

I ported the same fix and regression coverage onto the current `main` branch in #1457 so it can go through CI cleanly. This PR is effectively superseded by #1457.

hand-dot:
The ported fix has now been merged via #1457. Closing this PR as superseded. Thanks again for the original diagnosis and fix.

PR #1489, opened 2026-05-08
fix(common): handle same-y dynamic layout groups

EedgeY:
## Summary
- group horizontally aligned dynamic schemas by overlapping original Y ranges
- use group-level expansion offsets so same-Y siblings do not push each other down
- add coverage for same-Y, near-Y, multiple expansion, and page-spanning sibling cases

## Tests
- npm run test -w packages/common -- dynamicTemplate.test.ts

### DIFF
--- a/packages/common/__tests__/dynamicTemplate.test.ts
+++ b/packages/common/__tests__/dynamicTemplate.test.ts
@@ -528,4 +528,278 @@ describe('getDynamicTemplate', () => {
       expect(text!.position.y).toBe(60);
     });
   });
+
+  describe('Same Y position scenarios (horizontal layout)', () => {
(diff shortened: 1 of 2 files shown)


hand-dot:
I am not sure if the following issue was caused by the recent fix, but when using two lists, the UI rendering is incorrect.
The positional relationship between the two does not appear to be as intended.

Please let me know if you have any ideas about this.

PDF rendering is fine.

CleanShot 2026-05-08 at 16 20 59


[template.json](https://github.com/user-attachments/files/27510473/template.json)

EedgeY:
Ah, I hadn’t considered the subtraction case — sorry about that.
I’ll think about a better approach.

PR #1508, opened 2026-05-12
fix(generator): render each page's schemas in their own array order

mb-weberon:
## Summary

`generate` iterates each page through a globally-deduplicated schema-name array (`Set` across all pages) then resolves each name on the current page with `schemaPage.find(s => s.name == name)`. Two consequences:

1. **Z-order on a page no longer matches that page's own schema array order** if the same name appears on another page first. A background rectangle declared *after* an image on page 2 still draws *before* the image because the `Set` put the image at page-1's earlier index — flipping the intended z-order. Real-world incidence: an image field repeated on both faces of a postcard with the same field name vanishes on the back face beneath that face's solid-fill background rectangle.

2. **Multiple schemas with the same name on the same page silently drop all but the first** because `.find()` returns one match.

The fix iterates each page's schemas array directly. Render order = the page's own ordering. Names continue to flow through `input[name]` for non-readOnly schemas (`pdfme.generate` already collapses by name on the input side via `Set` — that contract is unchanged, only the per-page render loop is corrected).

The companion reproduction PR is #1509. This PR contains the same test code with the suite enabled (only explanatory comments, the suite title, and `describe.skip` differ). Fix review and regression coverage are tracked here; the standalone skipped-test PR can be closed without losing coverage. This PR remains open for conflict resolution and validation against current main.

## Test plan

- [x] `npm run lint` passes
- [x] `npm run fmt` passes
- [x] `npx vitest run __tests__/render-order.test.ts` — 2/2 pass post-fix
- [x] Same tests fail on `main` without the fix (verified by running them on the companion demo branch with `describe.skip` removed)
- [x] No regressions in the rest of the generator suite (pre-existing snapshot failures unaffected — they reproduce on `main` independently)
- [x] Tests use synthetic 4-color tiny PNGs at distinct widths (10×1, 50×1, 100×1, 500×1); no real-world templates or proprietary data

### DIFF
--- a/packages/generator/__tests__/render-order.test.ts
+++ b/packages/generator/__tests__/render-order.test.ts
@@ -1,0 +1,114 @@
+import generate from '../src/generate.js';
+import { Template, BLANK_A4_PDF, Schema } from '@pdfme/common';
+import { PDFDocument, PDFName } from '@pdfme/pdf-lib';
+import { image } from '@pdfme/schemas';
+
+const plugins = { image };
+
+// PNGs at distinct widths so the embedded XObject's IHDR-reported
+// width × height fingerprints which schema rendered it. Widths chosen to
+// be far apart so the data URLs differ in length and in their last-16
+// bytes — the image plugin caches PDFImage by
+// `${type}:${value.length}:${value.slice(0,16)}:${value.slice(-16)}`,
+// which would collide on near-identical small PNGs.
+const PNG_10x1 =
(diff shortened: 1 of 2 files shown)


hand-dot:
Superseded by https://github.com/pdfme/pdfme/pull/1626, which rebases this fix onto current `main` (preserving recent `generate.ts` improvements) and lands the per-page schema render-order correction plus regression tests.

Thanks @mb-weberon for identifying the bug, the fix direction, and the original tests.

PR #1510, opened 2026-05-12
feat: add Thai mixed-script rendering support in PDF generation

### LINKED ISSUE #1347: Thai tone mark (ไม้เอก / U+0E48) not visible when text starts with Latin character
### Describe the bug

I’m generating PDFs with **pdfme** and found a rendering issue with Thai combining marks (especially **ไม้เอก** `่` / `U+0E48`).

- If the text **starts with Thai**, the tone mark renders correctly.
- If the text **starts with Latin (English)** and later contains Thai, the tone mark may become **invisible / not rendered properly**.

Important detail: when I open the PDF and **copy/paste the affected text**, the pasted text still contains the correct characters (including the tone mark). So the Unicode text is present in the PDF, but the mark seems to be placed incorrectly during rendering.

### Steps to Reproduce
1. Create a template with one simple text field.
2. Render these two inputs into the same field:

**Case A (works):**
- `วันที่`

**Case B (bug):**
- `A วันที่` (or `ABC วันที่`)

3. Open the generated PDF in a viewer (browser PDF viewer / Adobe Reader).

### Expected Behavior
Thai tone marks like `่` (U+0E48) should be visible and positioned correctly in all cases, including when Thai text is preceded by Latin characters.

Image

Image

### Actual Behavior
In Case B, the tone mark appears missing. However:
- Copy/paste from the PDF returns the correct string, including the tone mark.
- This suggests the tone mark is not removed, but **rendered at the wrong position**.

### Additional Observation / Hypothesis
In the word `วันที่`, the character `ี` is a combining vowel mark (Sara I / U+0E35). My guess is that the tone mark `่` (U+0E48) is being **forced into the same position as `ี`**, causing overlap (so the tone mark is visually hidden behind/inside the vowel mark), instead of being shifted above as expected.

### Workarounds Tried
- Embedding a Thai font that includes Latin glyphs (Sarabun / Noto Sans Thai) and forcing all text fields to use a single `fontName` helps.
- Setting `subset: false` for the embedded font can also reduce the issue.

### Suspected Cause
Possibly related to **font fallback / font run segmentation** when the string begins with Latin text, which may affect mark-to-mark positioning (GPOS) for Thai combining marks.

### To Reproduce

I can provide a minimal reproducible repository / ZIP with:
- `package.json` (locks exact versions)
- a single script (e.g. `node 

.mjs`) that generates the PDF
- the Thai font file used (e.g. `Sarabun-Regular.ttf`) placed under `./fonts/`
- the generated output PDFs:
  - `output-thai-first.pdf` (works)
  - `output-latin-first.pdf` (bug)

[pdfme-thai-tone-mark-repro.zip](https://github.com/user-attachments/files/25384975/pdfme-thai-tone-mark-repro.zip)
 #### 1) Install
```bash
npm i @pdfme/generator @pdfme/common
# (optional) npm i -D tsx
```
### Expected behavior

- Thai combining marks (especially **ไม้เอก** `่` / `U+0E48`) should be **visible and positioned correctly** in all contexts.
- Rendering should be **consistent regardless of the first character/script** in the string (Thai-first vs Latin-first).
- Specifically, the following strings should render identically for the Thai portion:
  - `วันที่`
  - `A วันที่`
  - `ABC วันที่`
- In the word `วันที่`, the tone mark `่` (U+0E48) should be properly **stacked above** the base consonant and adjusted relative to other combining marks (e.g., `ี` / U+0E35), i.e. it should **not overlap or be placed at the same position** as the vowel mark.
- The visual rendering should match the underlying text layer: if copy/paste returns `วันที่`, the PDF should also visually display `วันที่` with the tone mark clearly visible.

### Your Environment

```markdown
- pdfme
  - "@pdfme/common": "^5.5.8",
  - "@pdfme/generator": "^5.0.0",
  - "@pdfme/schemas": "^5.0.0",
- Operating system: Ubuntu 24.08
- Node.js : 24
```

### Your Error Log

```shell
None Error
```

### Additional context

_No response_

pcbimon:
Fixes Thai mixed-script text rendering in PDF generation when Latin and Thai text are rendered in the same line.

Previously, strings such as `A วันที่` could lose or mis-render Thai tone marks during PDF rendering, even though the underlying copied text still contained the correct Unicode characters. This happened because mixed Latin/Thai text was passed to `page.drawText()` as a single text run.

This PR splits plain text into script runs before rendering, so Thai text is drawn separately from Latin text while preserving the visual position of the full line.

Fixes #1347

## Changes

- Implemented helper functions to split text into script runs for mixed Latin and Thai text.
- Enhanced the PDF rendering logic to render each script run separately.
- Preserved character spacing and line positioning across split runs.
- Applied rotation per run so rotated text still uses the correct pivot positioning.
- Added tests for verifying the correct segmentation of mixed-script text.
- Added a visual PDF smoke test for checking Thai tone marks in generated PDF output.
- Included `Sarabun-Regular.ttf` as a test asset for reliable Thai + Latin rendering in tests.

## Why

Thai tone marks require correct glyph shaping and positioning. When Latin and Thai characters are rendered together as a single text run, the PDF rendering stack can fail to position Thai combining marks correctly.

By splitting text into homogeneous script runs before calling `page.drawText()`, Thai shaping is isolated from Latin text and tone marks are preserved visually.

## Test cases covered

The new test cases include:

- `วันที่`
- `A วันที่`
- `AAA วันที่`
- `วันที่ A`
- `AAAAAA A วันที่ AA`
- `A วันที่ A วันที่ A`

These cover:

- Thai-only text
- Latin-before-Thai text
- Thai-before-Latin text
- Repeated mixed-script runs
- Mixed Latin/Thai text with surrounding spaces

## Testing

Run the mixed-script segmentation test:

```bash
npm run test -w packages/schemas -- thaiMixedScript.test.ts
````

Run the visual PDF smoke test:

```bash
npm run test -w packages/schemas -- thaiMixedScriptPdf.test.ts
```

The visual PDF smoke test generates:

```txt
packages/schemas/__tests__/output/thai-mixed-script-visual.pdf
```

Open the generated PDF and verify that:

* Latin characters render as normal letters, not tofu boxes.
* Thai text renders correctly.
* The tone mark in `วันที่` is visible in mixed-script lines such as `A วันที่` and `AAA วันที่`.

## Notes

The visual PDF test uses `Sarabun-Regular.ttf` because it supports both Thai and Latin glyphs. The font is only used as a test asset and does not affect runtime rendering behavior.

If `npm run typecheck` fails due to workspace linking or build order issues, build/link the workspaces first:

```bash
npm install
npm run link-workspaces
```

Then re-run the targeted tests:

```bash
npm run test -w packages/schemas -- thaiMixedScript.test.ts
npm run test -w packages/schemas -- thaiMixedScriptPdf.test.ts
```


hand-dot:
Thanks for the PR and for the clear repro around mixed Latin/Thai text (#1347).

After a deeper investigation on current `main`, we’ve decided **not to merge this approach**.

**What we found**
The root cause is in `@pdfme/pdf-lib`’s `encodeText` path: fontkit picks a **single** script for the whole string, and pdf-lib then emits glyph IDs via `Tj` while dropping GPOS `yOffset`. For Thai-only strings the GSUB path substitutes a pre-raised mark glyph, so marks still look fine; for Latin-leading strings (e.g. `A วันที่`) the mark stays on the default glyph and the needed vertical offset never reaches the PDF, so it collides with the vowel and “disappears”. The same class of failure also shows up with CJK-leading + Thai. Copy/paste still looks correct because of ToUnicode.

**Why this PR isn’t the long-term fix**
Splitting only at Latin/Thai boundaries in the schemas/PDF draw layer is a workable **mitigation** for some reported cases, but it doesn’t fix encoding for general mixed-script runs (e.g. CJK + Thai), and it leaves the underlying pdf-lib layout/encode bug in place. We want the durable fix to **script-itemize and call layout per run inside pdf-lib** (with proper script tags), rather than special-casing Thai in the higher layer.

We’ll track the proper fix against #1347. Really appreciate the investigation and the PR — it helped sharpen the diagnosis. Closing this PR for now; happy to review a follow-up aimed at the pdf-lib encode/layout path if you’re interested.

PR #1513, opened 2026-05-19
fix(ui): ensure updateOptions creates new options reference

marklawlor-payble:
## Summary

- `BaseUIClass.updateOptions()` was calling `Object.assign(this.options, newOptions)`, which mutates the existing options object in place.
- This means callers that hold a reference to the options object, or that compare options by reference (e.g. React Context value checks, reactive state systems), cannot detect that the options changed — the reference before and after the call is the same object.
- Fixed by changing to `Object.assign({}, this.options, options)` so each `updateOptions()` call assigns a fresh object to `this.options`, giving the new options a distinct reference.

## Test plan

- [ ] Existing test suite passes: `cd packages/ui && npm run test`
- [ ] New test `updateOptions creates a new options object reference` in `packages/ui/__tests__/class.test.tsx` verifies that `getOptions()` returns a different object reference after `updateOptions()` is called, and that the updated value is correctly reflected.
- [ ] `npm run lint` — no warnings or errors
- [ ] `npm run fmt` — no formatting changes needed

### DIFF
--- a/packages/ui/__tests__/class.test.tsx
+++ b/packages/ui/__tests__/class.test.tsx
@@ -13,6 +13,13 @@ class TestUI extends BaseUIClass {
   }
 }
 
+class ResizeObserverMock {
+  constructor(_callback: ResizeObserverCallback) {}
+  public observe() {}
+  public unobserve() {}
+  public disconnect() {}
+}
+
 const template: Template = {
   basePdf: BLANK_PDF,
   schemas: [[]],
@@ -20,14 +27,6 @@ const template: Template = {
 
 test('BaseUIClass mount renders without forcing a synchronous flush', async () => {
   const originalResizeObserver = globalThis.ResizeObserver;
-
-  class ResizeObserverMock {
-    constructor(_callback: ResizeObserverCallback) {}
-
-    public observe() {}
-    public unobserve() {}
-    public disconnect() {}
-  }
 
   globalThis.ResizeObserver = ResizeObserverMock as unknown as typeof ResizeObserver;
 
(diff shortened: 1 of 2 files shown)


marklawlor-payble (commented):
## Code Review

### Overview

This PR fixes a bug in `BaseUIClass.updateOptions()` where calling `Object.assign(this.options, newOptions)` mutated the existing options object in place. Any caller holding a reference to the old options (e.g. from `getOptions()`) would silently see the mutation, and reference-based change detection (React Context, MobX, reactive signals) could not detect the update.

The fix is correct and minimal.

---

### Code Quality

**`packages/ui/src/class.ts` line 131**

The one-line change is exactly right:
```ts
// Before (mutates in place — reference stays the same)
this.options = Object.assign(this.options, options);

// After (fresh object — new reference every call)
this.options = Object.assign({}, this.options, options);
```

---

### Suggestions

**1. Constructor has the same asymmetry (minor, out of scope but worth noting)**

At line 70:
```ts
this.options = options;
```
The constructor stores the caller's raw reference, so `getOptions()` initially returns the *same* object the caller passed in. After `updateOptions()` is called, it returns a fresh object. This asymmetry could surprise callers (e.g. `getOptions() === passedOptions` is true until the first `updateOptions()`, then false). Defensively shallow-copying in the constructor too would make the class consistent:
```ts
this.options = { ...options };
```
Not a blocker, but worth a follow-up.

**2. Test: consider asserting the old reference is unchanged**

The test verifies the new reference is different and has the updated value, which is sufficient. Adding one more assertion would make the immutability guarantee explicit:
```ts
expect(optionsBefore.lang).toBe('en'); // original not mutated
```

---

### Test Coverage

- The `ResizeObserverMock` refactor (moving it to module scope) is a clean improvement — removes repetition without altering behaviour.
- The new test covers the happy path correctly: different reference after update, correct merged value.
- Setup/teardown mirrors the existing test pattern (try/finally, DOM cleanup, ResizeObserver restore).

---

### Summary

| Area | Assessment |
|---|---|
| Correctness | Fix is correct |
| Scope | Appropriately narrow |
| Test | Good coverage of the bug |
| Style | Matches project conventions |

The fix is ready to merge. The constructor asymmetry noted above is a separate concern and doesn't block this PR.

marklawlor-payble:
@greptile-apps review again

hand-dot:
Superseded by #1618 (same shallow-copy fix rebased onto current `main`). Closing in favor of that PR — thanks for the original fix and tests.

Issue #1235, opened 2025-11-19
Add optional page field to updateTemplate() so it doesn't always reset to page 0


akhil-bp:
### Is your feature request related to a problem? Please describe.

Currently, when calling **updateTemplate()** in pdfme, the active page always resets to page 0.
This becomes an issue when adding fields programmatically to a different page. The field gets added correctly to the desired page, but the viewer automatically jumps back to the first page, causing you to lose context while editing.

### Describe the solution you'd like

**Currently, when calling updateTemplate() in pdfme, the active page always resets to page 0.**
This becomes an issue when adding fields programmatically to a different page. The field gets added correctly to the desired page, but the viewer automatically jumps back to the first page, causing you to lose context while editing.

Requesting support for an optional page parameter in updateTemplate() so that the caller can control which page to stay on or switch to after the update.

**Steps to Reproduce**
- Open a multi-page template.
- Programmatically add a field to a non-zero page (e.g., page 2).
- updateTemplate() is internally called to apply the changes.
- After the update, the viewer always jumps back to page 0.

**Proposed Enhancement**
Add an optional page argument to updateTemplate():
Eg : designer.current.updateTemplate(template, { page: 2 });

Genzo0:
any updates on this ?

superiority-claw-agent:
Fixed in [PDFweave](https://github.com/IDNTEQ/pdfweave) — [PR #1](https://github.com/IDNTEQ/pdfweave/pull/1), commit [`1e1f0d14`](https://github.com/IDNTEQ/pdfweave/commit/1e1f0d14). Will ship in @pdfweave/* v0.1.1.

Happy to send a PR upstream if a maintainer wants to take it — the diff is isolated. Context on the PDFweave fork: pdfme/pdfme#1468.

— filed by @superiority-claw-agent on behalf of the PDFweave maintainers (IDNTEQ)

hand-dot:
We checked the reported page reset with @pdfme/ui 6.1.12: on a two-page template, adding a field to the second page through updateTemplate() preserved both page index 1 and the canvas scroll position (777 before and after).

The requested optional page argument is a separate API enhancement and is not implemented by that result. selectSchemas() now supports a pageIndex option through #1528, which may cover selecting a newly added field on a target page.

[...]

PR #1515, opened 2026-05-19
fix(ui): preserve scroll position when updating template

marklawlor-payble:
## Problem

When `Designer.updateTemplate()` is called on a multi-page template, the canvas scrollTop resets to 0, snapping the user back to page 1. This is a significant UX regression — users editing page 3 are involuntarily sent back to the top after any autosave or sidebar-triggered template update.

## Root Cause

`TemplateEditor` (the React component inside Designer) maintains three separate async state slices that describe the current template:

- `schemasList` — updated by `template2SchemasList()` (async)
- `pageSizes` and `backgrounds` — updated by `useUIPreProcessor`'s `runInit()` (async)

`Paper.tsx` contains a guard:

```tsx
if (pageSizes.length !== backgrounds.length || pageSizes.length !== schemasList.length) {
  return null;
}
```

When a new `template` prop arrives, these three slices update in separate React render passes. During the brief window where they are mismatched, `Paper` returns `null`, collapsing the scrollable content height to zero. The browser immediately clamps `scrollTop` to 0 (the new maximum), and it never recovers.

## Fix

1. **Save `scrollTop` before the update.** At the start of `updateTemplate` when `preservePage=true`, capture `canvasRef.current.scrollTop` into a ref (`scrollRestoreRef`).

2. **Restore once state settles.** A new `useEffect` watches `[pageSizes, backgrounds, schemasList]`. When all three are non-empty and have equal length (the transient mismatch has resolved), it writes the saved value back to `scrollTop` and clears the ref.

This approach is robust: it doesn't rely on `setTimeout`, `requestAnimationFrame`, or any assumed timing — it restores exactly when the DOM is ready.

The `!preservePage` branch (external API `updateTemplate` with no scroll preservation) remains unchanged — it explicitly scrolls to the top.

## Test

A new test in `packages/ui/__tests__/components/Designer.test.tsx` verifies the mechanism.

Because jsdom provides no real scroll layout (`scrollTop` always reads 0), the test uses `Object.defineProperty` to override `scrollTop` on the canvas element with a custom getter/setter that simulates being at position 500px. After re-rendering with an updated template, it asserts that the restore effect has written 500 back to `scrollTop`.

## Changes

- `packages/ui/src/components/Designer/index.tsx` — fix + `scrollRestoreRef`
- `packages/ui/__tests__/components/Designer.test.tsx` — new regression test

## Issue linkage

Related to #1235. This changes scroll restoration and internal page-update handling. It does not add the requested public updateTemplate(template, { page }) API, so it should not automatically close that broader request. Validate the original external updateTemplate sequence separately.

### DIFF
--- a/packages/ui/__tests__/components/Designer.test.tsx
+++ b/packages/ui/__tests__/components/Designer.test.tsx
@@ -72,23 +72,12 @@ test('Designer keeps toolbar zoom interactive when options.zoomLevel is only an 
 });
 
 test('Designer restores scroll position when updateTemplate is called while scrolled', async () => {
-  // jsdom does not implement layout so scrollTop always reads 0.
-  // We verify the scroll-restore mechanism by overriding scrollTop on the canvas
-  // element so that it reports a non-zero value, then checking it is written back
-  // to that value after the template prop changes.
-  //
-  // Root cause: Paper returns null during the transient window where pageSizes,
-  // backgrounds, and schemasList are out of sync after a template update. The
-  // browser collapses scrollable height to 0 and clamps scrollTop. The fix saves
-  // scrollTop before the update and restores it once all three collections are
-  // back in sync.
-  //
-  // Note: because jsdom has no real layout engine, we use Object.defineProperty to
-  // simulate a non-zero scrollTop and verify that the restore logic writes it back.
+  // jsdom has no layout engine so we use Object.defineProperty to simulate a
+  // non-zero scrollTop and verify the restore effect writes it back.
   setupUIMock();
 
   const template = getSampleTemplate();
-  let currentScrollTop = 500; // simulate being scrolled into a later page
+  let currentScrollTop = 500;
   const scrollTopSetter = vi.fn((v: number) => {
     currentScrollTop = v;
   });
@@ -114,7 +103,6 @@ test('Designer restores scroll position when updateTemplate is called while scro
   const canvas = container.querySelector('.pdfme-designer-canvas') as HTMLDivElement;
   expect(canvas).not.toBeNull();
 
-  // Override scrollTop to simulate being scrolled 500px into the canvas
   Object.defineProperty(canvas, 'scrollTop', {
     get: () => currentScrollTop,
     set: scrollTopSetter,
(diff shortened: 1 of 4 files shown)


marklawlor-payble:
Addressed the feedback from the Greptile review:

**Problem**: `updatePage` (called by add/remove page) called `updateTemplate(preservePage=true)`, which saved the current scroll position into `scrollRestoreRef`. The restore effect then wrote that old-page offset back to the canvas before `updatePage`'s own `setTimeout(0)` could set the correct new-page position — producing a visible scroll flash on every page add/remove.

**Changes** (`f7a1cb5`):

1. **`preserveScroll` parameter on `updateTemplate`** — added a third parameter `preserveScroll` (defaults to `preservePage`). `updatePage` now passes `false`, so the restore mechanism is completely skipped on those code paths. External autosave callers continue to use the default (`preserveScroll=true`).

2. **Strict Mode purity fix** — moved the `scrollRestoreRef.current` mutation out of the `setPageCursor` state updater callback. React Strict Mode invokes updater functions twice to surface side effects; a ref write inside an updater is exactly the kind of impurity that causes double-fires. The clamped page index is now computed outside and `setPageCursor` receives the value directly.

3. **Regression test** — added `'Designer does not flash old scroll position when adding a page'` which renders a blank-PDF template (required to enable page manipulation), mocks `scrollTop` to 500, clicks the "Add Page After" menu item, and asserts `scrollTopSetter` was never called with `500`.

4. **`ResizeObserver` polyfill in `vitest.setup.ts`** — jsdom doesn't provide `ResizeObserver`; Ant Design's Dropdown and `@dnd-kit` require it when a menu is opened. Added a no-op class polyfill so the test (and any future tests that open dropdowns) don't crash.

marklawlor-payble:
@greptile-apps  review again

marklawlor-payble:
@greptile-apps review again

hand-dot:
Thanks for the PR to preserve scroll position across `updateTemplate`.

We re-checked the reported “reset to page 0” behavior on current `main` (Designer, multi-page, `updateTemplate` while on page 2). **We could not reproduce the page reset** — page cursor and scroll stay put. That looks covered by the earlier `preservePage` work (#1431).

So we’re going to **close this PR** as no longer needed for the original bug. The branch is also dirty against current `main`.

If you still need an explicit `{ page }` argument on `updateTemplate` as an enhancement, let’s keep that discussion on #1235 (or a fresh Issue) rather than this scroll-restore approach. Appreciate the contribution.

Issue #263, opened 2023-09-13
Allow more accurate schema placement by supporting positioning relative to guides/rulers


peteward:
### Is your feature request related to a problem? Please describe.

When working towards precise designs it can be difficult to place items exactly as you want them.

This is usually when you want to align something with respect to other static items in the PDF, rather than the whole PDF itself.  You might want to place a text box equally spaced between 2 other paragraphs, or centred with respect to another container (but not as wide as it).

### Describe the solution you'd like

Allow the use of guides as a point of reference to allow positioning. 

Roughly the idea would be that if you are positioning an item in a dimension for which there are one or more guides, the positioning will be done with respect to those guides rather than the whole page.

e.g.:

If I wanted to center the schema horizontally between these 2 guides I could hit the horizontally centering icon to snap it into position:
![Screenshot 2023-09-13 at 14 01 07](https://github.com/pdfme/pdfme/assets/7068515/3e1c3eb5-165f-4f4a-98e6-78febf9edecc)

Similarly, you could align to the sides if there's a single rule (or to the nearest rule).

The exact behaviour for each positional action needs more thought before continuing.

### Describe alternatives you've considered

Some times you can grow the item to the size of the container, but this doesn't work in many cases.
You can also use rulers and do some math on the numbers to place items with more precision, but this isn't very user friendly.

I'm not sure what other implementations could be used here, I think using the guides makes sense if it's technically viable.

superiority-claw-agent:
PDFweave's anchor layouts (in @pdfweave/common — `HorizontalAnchorRule` / `VerticalAnchorRule`) cover guided positioning by binding x/y to anchor references on other schemas. See `packages/common/src/types.ts`.

If a maintainer wants this surfaced upstream too, happy to send a PR — but the work is already in PDFweave at v0.1.0. Context on the fork: pdfme/pdfme#1468.

— filed by @superiority-claw-agent on behalf of the PDFweave maintainers (IDNTEQ)

hand-dot:
Linked #1516 as related work: it exposes getGuides/setGuides for reading and restoring guide positions. This is useful guide infrastructure, but it does not implement centering or alignment relative to guides as requested here. This issue remains open for that behavior.

hand-dot:
Maintainer triage update (current `main` checked at `7501849e`): status is `partially_done`.

[...]

PR #1516, opened 2026-05-19
feat(ui): add getGuides and setGuides public methods to Designer

marklawlor-payble:
## Problem

The `Designer` class had no public API to read or write guide positions. Applications that needed to save/restore guide state (for example, persisting alignment rulers across sessions) had to walk the React fiber tree to find `@scena/react-guides` component instances by scanning for `__reactFiber$`-prefixed keys on DOM nodes. That technique is fragile and breaks whenever the bundled React version changes its internal key scheme.

## Solution

This PR adds two public methods to the `Designer` class:

```ts
type Guides = { horizontal: number[][]; vertical: number[][] }

designer.getGuides(): Guides
designer.setGuides(guides: Guides): void
```

The shape is multi-page — the outer array index corresponds to the page, the inner array holds guide positions (in the same unit as the ruler, consistent with `@scena/react-guides`). This matches the `template.schemas: SchemaForUI[][]` multi-page convention already used throughout the codebase.

`Guides` is also exported from `@pdfme/ui` for consumers who need to type-annotate saved state.

## Why not `setOptions`?

Guides could have been tucked into `updateOptions({ guides: ... })` but that would be the wrong fit for three reasons:

1. **`UIOptions` is shared across `Viewer`, `Form`, and `Designer`** — guides are Designer-only. Adding them to the shared Zod schema in `@pdfme/common` would leak a Designer-specific concern into the base class.

2. **Wrong merge semantics.** `updateOptions` does a shallow `Object.assign` merge. Guides are `number[][]` per page; there is no natural "merge" — a partial update would silently replace the whole array in a surprising way. A dedicated `setGuides` is unambiguous.

3. **Precedent.** `getTemplate` / `setTemplate` are already standalone methods for imperative state even though `template` is passed at construction. Guides are per-page user-positioned data — closer in kind to template state than to display configuration — so they follow the same pattern.

## Usage example

```ts
import { Designer, type Guides } from '@pdfme/ui';

const designer = new Designer({ domContainer, template });

// Save guides
const saved: Guides = designer.getGuides();
localStorage.setItem('guides', JSON.stringify(saved));

// Restore guides
const restored: Guides = JSON.parse(localStorage.getItem('guides') ?? '{}');
designer.setGuides(restored);
```

## Implementation

The wiring uses a controller-object pattern that mirrors the existing `onPageCursorChange` callback:

1. `Canvas/index.tsx` — defines `GuidesController` (with `getGuides` / `setGuides` that delegate to the per-page `horizontalGuides` / `verticalGuides` refs), and calls `onMountGuidesController` once after mount.
2. `Designer/index.tsx` (TemplateEditor) — accepts and threads `onMountGuidesController` down to `Canvas`.
3. `Designer.tsx` (the public class) — stores the controller in `this.guidesController` and exposes the two public methods. `getGuides()` returns empty arrays per page if called before mount. `setGuides()` silently no-ops for page indices that have no mounted ref.

## Tests

Three new tests in `__tests__/components/DesignerGuides.test.tsx` (using a `vi.mock` of `@scena/react-guides` with a stub that honours `getGuides` / `loadGuides`):

- `getGuides` returns an empty array per page when no guides have been set
- `setGuides` followed by `getGuides` round-trips correctly
- `setGuides` with more pages than are mounted does not throw

All 59 tests pass (`npm run test -w packages/ui`). Lint and format are clean.

## Issue linkage

Related to #263 as supporting guide infrastructure. getGuides/setGuides exposes guide positions for reading and restoration; it does not implement the requested alignment/centering actions relative to one or two guides. This is not a complete fix for that issue.

### DIFF
--- a/packages/ui/__tests__/components/DesignerGuides.test.tsx
+++ b/packages/ui/__tests__/components/DesignerGuides.test.tsx
@@ -1,0 +1,126 @@
+import React from 'react';
+import { render, act, waitFor } from '@testing-library/react';
+import { vi } from 'vitest';
+import Designer from '../../src/components/Designer/index.js';
+import { type GuidesController } from '../../src/components/Designer/Canvas/index.js';
+import { I18nContext, FontContext, PluginsRegistry } from '../../src/contexts';
+import { i18n } from '../../src/i18n';
+import { getDefaultFont, pluginRegistry } from '@pdfme/common';
+import { setupUIMock, getSampleTemplate } from '../assets/helper';
+import { text, image } from '@pdfme/schemas';
+import { SELECTABLE_CLASSNAME } from '../../src/constants';
+
+const plugins = { text, image };
+
+/**
+ * Stub GuidesComponent so that in jsdom we can control getGuides / loadGuides
+ * without needing real canvas layout.  Each instance stores its own guide array.
+ */
+const makeGuideInstance = () => {
+  let guides: number[] = [];
+  return {
+    getGuides: () => guides,
+    loadGuides: (g: number[]) => {
+      guides = g;
+    },
+    scroll: () => {},
+    scrollGuides: () => {},
+    resize: () => {},
+  };
+};
+
+// Map from the underlying DOM element to our stub instance so that forwardRef
+// can produce a stable ref object.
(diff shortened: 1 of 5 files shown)


marklawlor-payble:
@greptile-apps  review again

PR #1517, opened 2026-05-19
feat(ui): add selectSchemas public method to Designer

marklawlor-payble:
## Problem

The `Designer` class has no public API to programmatically select schema elements. Applications that need to auto-select a newly placed element — for example, after a drag-and-drop from an external palette — currently have no supported path. The only workaround is to find the internal `.selectable` DOM node and synthesise a click/mouse event, which is fragile and breaks across pdfme versions.

## Solution

Add a `selectSchemas(names: string[])` public method to `Designer` that selects one or more elements by their schema `name` property.

### Method signature

```ts
designer.selectSchemas(names: string[])
```

- **`names`** – array of schema names to select. Names that don't match any schema in the template are silently ignored.
- Passing an empty array deselects everything (equivalent to clicking an empty area of the canvas).
- Only schemas on the currently-visible page can be targeted, because the canvas only renders DOM nodes for the active page. If the target schema is on a different page, navigate there first.

### Usage example

```ts
import { Designer } from '@pdfme/ui';

const designer = new Designer({ domContainer, template, plugins });

// After programmatically adding a schema, auto-select it:
designer.onChangeTemplate((t) => {
  const lastSchema = t.schemas[0].at(-1);
  if (lastSchema) {
    designer.selectSchemas([lastSchema.name]);
  }
});

// Select multiple schemas at once:
designer.selectSchemas(['firstName', 'lastName']);

// Deselect everything:
designer.selectSchemas([]);
```

## Implementation

The selection state (`activeElements` / `onEdit`) lives inside the `TemplateEditor` functional component and is not accessible from the `Designer` class directly. The bridge is a callback-registration pattern consistent with the existing `onSaveTemplate` / `onChangeTemplate` style:

1. `TemplateEditor` gains an optional `onMountEditorApi` prop.
2. On mount it calls `onMountEditorApi` with a `DesignerEditorApi` object `{ selectSchemas }`.
3. `Designer` stores the API handle in a private `editorApi` field and proxies `selectSchemas` through it.
4. A `schemasListRef` keeps the closure over `schemasList` fresh so the API handle never becomes stale.
5. A `setTimeout` (matching the existing `addSchema` pattern) ensures DOM nodes exist before calling `onEdit`.

## Test

A new test in `__tests__/components/Designer.test.tsx` renders the component, calls `selectSchemas(['field1'])`, and asserts that the canvas delete button (`pdfme-designer-delete-button`) appears — the same visual cue that appears when a user clicks a schema. It then calls `selectSchemas([])` and asserts the button disappears.

## Files changed

- `packages/ui/src/components/Designer/index.tsx` — `DesignerEditorApi` type + `onMountEditorApi` prop + `useEffect` bridge
- `packages/ui/src/Designer.tsx` — `editorApi` field + `selectSchemas` public method + wires `onMountEditorApi`
- `packages/ui/__tests__/components/Designer.test.tsx` — new test

### DIFF
--- a/packages/ui/__tests__/components/Designer.test.tsx
+++ b/packages/ui/__tests__/components/Designer.test.tsx
@@ -1,6 +1,7 @@
 import React from 'react';
 import { render, act, fireEvent, waitFor } from '@testing-library/react';
-import Designer, { type DesignerEditorApi } from '../../src/components/Designer/index.js';
+import Designer from '../../src/components/Designer/index.js';
+import DesignerClass from '../../src/Designer.js';
 import { I18nContext, FontContext, OptionsContext, PluginsRegistry } from '../../src/contexts';
 import { i18n } from '../../src/i18n';
 import { DESIGNER_CLASSNAME, RIGHT_SIDEBAR_WIDTH, SELECTABLE_CLASSNAME } from '../../src/constants';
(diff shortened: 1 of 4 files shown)


marklawlor-payble:
Addressed all feedback from the review:

**Greptile P1 — Silent no-op before mount** (commit 6709f61)
Replaced the silent null-guard with an explicit throw:
```ts
throw Error(
  '[@pdfme/ui] selectSchemas was called before the Designer finished mounting. ' +
  'Wait for the component to mount (e.g. inside onChangeTemplate) before calling this method.'
);
```
This matches the throw-or-succeed convention used by every other public method on the class.

**Greptile P2 — Test covers internal API, not public class** (commit 6709f61)
Replaced the `DesignerEditorApi`-prop component test with two tests that instantiate `new Designer(...)` directly:
- Verifies `selectSchemas` selects and deselects schemas via the public class method (post-mount path)
- Verifies `selectSchemas` throws the descriptive error when called before the component has mounted

**Greptile P2 — `document.getElementById` scope** (commit 6709f61)
Scoped the DOM query to `canvasRef.current` first, falling back to `document`:
```ts
const root = canvasRef.current ?? document;
const elements = matchedIds
  .map((id) => root.querySelector<HTMLElement>(`[id="${id}"]`))
  .filter((el): el is HTMLElement => el !== null);
```

**Review — `DesignerEditorApi` not exported from package barrel** (commit 0ab37781)
Re-exported from `packages/ui/src/index.ts` so TypeScript consumers can import it directly:
```ts
import type { DesignerEditorApi } from '@pdfme/ui';
```

**Review — Misleading comment** (commit 0ab37781)
Fixed the comment on `schemasListRef.current.flat()` — it searches all pages by name, not the current page first. The DOM lookup inside the `setTimeout` is still page-scoped via `canvasRef`.

marklawlor-payble:
@greptile-apps review again

marklawlor-payble:
Addressed the timing inconsistency flagged in the review.

**Change:** Moved the \`names.length === 0\` early-return inside the \`setTimeout\` callback so both the deselect path and the select path share the same async timing contract. Previously, calling \`selectSchemas([])\` fired \`onEditEnd\` synchronously while \`selectSchemas(['x'])\` queued \`onEdit\` via \`setTimeout\`, meaning back-to-back calls in the same tick would execute in the wrong order.

@greptile-apps — the suggestion from review [#discussion_r3264349268](https://github.com/pdfme/pdfme/pull/1517#discussion_r3264349268) has been applied.

marklawlor-payble:
@greptile-apps review again

marklawlor-payble:
Addressed the remaining greptile P1 issue — the unmatched-name deselect bug.

**Fix (commit 4b6b87ef):** Added `if (elements.length === 0) return;` guard before calling `onEdit(elements)` in the `selectSchemas` closure. Previously, when every name in the array was either absent from the schema list or had no DOM node on the current page, `elements` resolved to `[]` and `onEdit([])` silently cleared the active selection — contradicting the documented "silently ignored" contract.

**Test:** Added a regression test that selects `field1`, then calls `selectSchemas(['this_name_does_not_exist'])`, and asserts the delete button is still visible (selection preserved).

All 6 Designer tests pass.

marklawlor-payble:
@greptile-apps review again

marklawlor-payble:
Super-seeded by https://github.com/pdfme/pdfme/pull/1528

PR #1519, opened 2026-05-19
fix(common): support hyphenated field names in template expressions

marklawlor-payble:
## Problem

Template expressions that reference field names containing hyphens (e.g. `{my-field}` or `{first-name + " " + last-name}`) were silently broken.

**Root cause:** The expression evaluator in `packages/common/src/expression.ts` passes the inner content of every `{…}` placeholder to Acorn for JavaScript AST parsing. Since a hyphen is not a valid identifier character in JavaScript, Acorn parses `field-name` as a *subtraction expression* (`field − name`) rather than an identifier. When evaluation then tries to resolve `field` and `name` as separate context keys, neither exists (the real key is `field-name`), so the placeholder is left unreplaced in the output.

**Silent-corruption variant:** If the context happens to contain numeric fields named `field` and `name`, the expression evaluates as arithmetic subtraction and produces a number instead of the intended field value.

**Compound expressions** (e.g. `{first-name + " " + last-name}`) were completely unsupported: even with the simple key look-up fix, any expression that *combined* a hyphenated name with an operator still fell through to the Acorn parser, which mis-parsed the hyphen.

### Whole-placeholder fast-path

When the entire placeholder is a literal context key (e.g. `{my-field}`), the value is returned directly before Acorn is ever invoked. This handles the common case with zero tokenisation overhead.

### Tokenizer-based preprocessing for compound expressions

For expressions that *use* a hyphenated key alongside operators (e.g. `{first-name + " " + last-name}`), a preprocessing step runs the Acorn tokenizer over the expression and finds adjacent identifier-hyphen-identifier chains whose concatenated form is a known context key. Each matching span is rewritten to a safe JS identifier (`__k_first_u2d_name__`) and a corresponding entry is added to the evaluation context. Acorn then parses the clean expression normally.

Using the tokenizer (rather than naive string replacement) means string literals and whitespace-separated operators are never touched — `{name + "-suffix"}` and `{a - b}` remain correct.

**Disambiguation rule:** `{a-b}` resolves to the field named `a-b` if that key exists in context. To force arithmetic subtraction, use whitespace: `{a - b}`.

## Examples

```
// Template field content: "Hello, {first-name}!"
// Input variables: { "first-name": "Alice" }
// → "Hello, Alice!"

// Template field content: "{first-name + " " + last-name}"
// Input variables: { "first-name": "John", "last-name": "Doe" }
// → "John Doe"

// Template field content: "{nick-name || first-name}"
// Input variables: { "nick-name": "", "first-name": "Alice" }
// → "Alice"

// Template field content: "{is-active ? "yes" : "no"}"
// Input variables: { "is-active": true }
// → "yes"
```

## Test plan

- [x] Standalone hyphenated field: `{field-name}` → field value
- [x] Multiple standalone hyphenated fields in one template
- [x] Whitespace inside braces: `{ field-name }` (trimming still works)
- [x] Schema source for hyphenated names
- [x] Compound: string concatenation `{first-name + " " + last-name}`
- [x] Compound: arithmetic `{count-down + 1}`
- [x] Compound: logical OR fallback `{nick-name || first-name}`
- [x] Compound: ternary `{is-active ? "yes" : "no"}`
- [x] String literal with hyphen not corrupted: `{first-name + "-" + last-name}` → `"John-Doe"`
- [x] Disambiguation: `{a-b}` with key `a-b` resolves to the key, not subtraction
- [x] Disambiguation: `{a - b}` (spaced) evaluates as subtraction even when `a-b` exists in context
- [x] Regression: `{a-b}` with separate `a` and `b` keys (no `a-b` key) still subtracts
- [x] Regression: `{1 + 2}` → `3`
- [x] Regression: `{Math.max(1,2,3)}` → `3`
- [x] All existing expression tests pass

### DIFF
--- a/packages/common/__tests__/expression.test.ts
+++ b/packages/common/__tests__/expression.test.ts
@@ -297,6 +297,29 @@ describe('replacePlaceholders', () => {
     const result = replacePlaceholders({ content, variables: {}, schemas });
     expect(result).toBe('Content: ');
   });
+
+  it('should not collide when a context key matches a safe identifier name', () => {
+    // Regression for toSafeIdentifier collision: an older encoding produced a safe id
+    // that could be identical to a literal context key (e.g. __k_firstu2dname__ for
+    // both "first-name" and a key literally named "__k_firstu2dname__"). The index-based
+    // scheme produces __pdfme_hk_0__ etc., which cannot appear in user-supplied keys
+    // without quoting and so is collision-free in practice.
+    const variables = { 'first-name': 'John', '__pdfme_hk_0__': 'WRONG' };
+    const content = 'Hello {first-name}';
+    const result = replacePlaceholders({ content, variables, schemas: [] });
+    expect(result).toBe('Hello John');
+  });
+
+  it('should evaluate {1+1} as arithmetic even when "1+1" is a context key', () => {
+    // Regression for over-broad fast-path: previously any context key matching
+    // the placeholder content short-circuited evaluation, so {1+1} with a context
+    // key "1+1" would return the key value instead of 2. Narrowing the fast-path
+    // to non-identifiers only restores correct arithmetic behaviour.
+    const variables = { '1+1': 'WRONG' };
+    const content = 'Result: {1+1}';
+    const result = replacePlaceholders({ content, variables, schemas: [] });
+    expect(result).toBe('Result: 2');
+  });
 });
 
 describe('replacePlaceholders - Security Tests', () => {
--- a/packages/common/__tests__/expression.test.ts
+++ b/packages/common/__tests__/expression.test.ts
@@ -274,6 +274,29 @@ describe('replacePlaceholders', () => {
     const result = replacePlaceholders({ content, variables: {}, schemas: [] });
     expect(result).toBe('Max: 3');
   });
+
+  it('should resolve a three-part hyphenated key {a-b-c}', () => {
+    const content = 'Value: {a-b-c}';
+    const variables = { 'a-b-c': 'three-parts' };
+    const result = replacePlaceholders({ content, variables, schemas: [] });
+    expect(result).toBe('Value: three-parts');
+  });
+
+  it('should use empty string for a non-readOnly schema field with a hyphenated name', () => {
+    const content = 'Content: {my-field}';
+    const schemas = [
+      [
+        {
+          name: 'my-field',
+          type: 'text',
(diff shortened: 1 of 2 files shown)


marklawlor-payble:
Addressed the Greptile review feedback (comment #4484058167):

**1. `toSafeIdentifier` collision risk → fixed**
Replaced the hex-encoding scheme (`__k_firstu2dname__`) with a sequential index scheme (`__pdfme_hk_0__`, `__pdfme_hk_1__`, …). The old encoding could collide: a context key literally named `firstu2dname` would map to the same safe identifier as `first-name`. The new scheme assigns indices in left-to-right expression order and re-uses the same index for repeated occurrences of the same key, so it's guaranteed collision-free. A regression test covers the collision case.

**2. Over-broad fast-path → narrowed**
The original fast-path short-circuited on *any* context key match, so `{1+1}` with a context key `"1+1"` would return the key value instead of evaluating to `2`. Changed the guard to `isHyphenatedIdentifier(code)` — a regex that matches only `word-word[-word…]` patterns — so the fast-path fires only for the case it was designed for (hyphenated field names that Acorn would misparse as subtraction). A regression test verifies `{1+1}` still evaluates as arithmetic.

**3. Cache accumulation** — Greptile flagged this as "correct", no action taken.

**Bonus: pre-existing TypeScript error fixed**
The acorn `Token` type declaration omits the runtime `value` field, causing a `TS2339` error in `tsc -p tsconfig.build.json`. Added a local cast (`AcornTokenWithValue`) to fix it.

marklawlor-payble:
@greptile-apps  review again

hand-dot:
Thanks for the careful work on this — the problem is real, and the PR is thoughtfully tested.

We're going to close it without merging for now. Template expressions are intentionally evaluated as JavaScript, so identifiers follow JS rules (hyphens are subtraction, not part of a name). Supporting hyphenated keys inside `{…}` adds lasting parser edge cases (key vs subtraction, member access like `obj.first-name`, safe-placeholder collisions) that we don't want to take on while demand for this remains low.

Workarounds for callers today:
- Prefer JS-safe field names in expressions (e.g. `firstName` / `first_name`)
- Or keep hyphenated keys only as plain data bindings outside compound expressions, if your integration allows

If more users hit this and we decide to support it later, we can reopen or revisit with a narrower design (for example, whole-placeholder key lookup only). Appreciate the contribution.

PR #1553, opened 2026-06-12
Fix ReferenceError: bwipp_setanycolor is not defined when barcode includeText is true by adding a global no-op fallback.

### LINKED ISSUE #1427: ReferenceError: bwipp_setanycolor is not defined
### Describe the bug

This is a wonderful library to work with and I have just upgraded to latest version of this package (v6.0.6), so while using the barcode component, following error is thrown
"ReferenceError: bwipp_setanycolor is not defined"

This is because "includeText" flag holds true value.  Attaching the screenshot of the same as follows:

Image

Could you please help resolve this issue?

### To Reproduce

Go to examples, follow this [link](https://pdfme.com/template-design)
And drag and drop the 2nd last option, shown below:

Image

### Expected behavior

It should display the barcode text as provided by the user

### Your Environment

```markdown
- pdfme package(@pdfme/generator or @pdfme/ui): 6.0.6
- pdfme version: 6.0.6
- Operating system: macOS ventura
- Node.js version or Browser name & version: 18.16.0
```

### Your Error Log

```shell
ReferenceError: bwipp_setanycolor is not defined
```

### Additional context

_No response_

Federicorao:
Fixes #1427.

Fix ReferenceError: bwipp_setanycolor is not defined when barcode includeText is true by adding a global no-op fallback. BWIPP's barcode rendering calls bwipp_setanycolor when includeText is true, but the function is not defined in the global scope. Adding a check and a no-op function prevents the ReferenceError.

I ran the available lightweight check for the frontend change.

### DIFF
--- a/packages/schemas/src/barcodes/helper.ts
+++ b/packages/schemas/src/barcodes/helper.ts
@@ -127,6 +127,13 @@ export const barCodeType2Bcid = (type: BarcodeTypes) =>
 export const mapHexColorForBwipJsLib = (color: string | undefined, fallback?: string) =>
   color ? color.replace('#', '') : fallback ? fallback.replace('#', '') : '000000';
 
+const ensureBwippSetAnyColor = () => {
+  const globalScope = globalThis as typeof globalThis & { bwipp_setanycolor?: () => void };
+  if (typeof globalScope.bwipp_setanycolor !== 'function') {
+    globalScope.bwipp_setanycolor = () => {};
+  }
+};
+
 export const createBarCode = async (arg: {
   type: BarcodeTypes;
   input: string;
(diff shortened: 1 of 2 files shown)


Federicorao:
Updated the PR to address the Greptile finding.

- Restored `packages/schemas/src/barcodes/pdfRender.ts` to the original named-export implementation instead of the broken stub.
- Moved the minimal `bwipp_setanycolor` fallback into `createBarCode` in `helper.ts`, only when `includetext` is enabled.

Validation run locally:
- `npm ci`
- `npm run lint --workspace @pdfme/schemas`
- `npm run build --workspace @pdfme/pdf-lib && npm run build --workspace @pdfme/common && npm run build --workspace @pdfme/schemas`
- `npm run test --workspace @pdfme/schemas` (224 passed)

The remaining Vercel check still requires authorization by a member of the `labelmake` Vercel team.

hand-dot:
Thanks for the contribution and for addressing the initial review feedback. Closing this PR without merging because the global no-op workaround is no longer needed for the verified current-release case.

The original issue #1427 has a user report confirming resolution in 6.1.2. I additionally verified the published 6.1.12 packages on Node.js 24.14.0: Code 128 and EAN-13 PDF generation with `includetext: true` succeeds without defining `globalThis.bwipp_setanycolor`, and the rendered bars and text are present.

We should avoid introducing a global no-op for a problem that no longer reproduces in this check. If a current browser/bundler-specific reproduction remains, please share it in #1427 so we can investigate the actual failing path.

Issue #1561, opened 2026-06-24
Clicking Color Picker Causes White Screen


necessarylion:
### Describe the bug

Clicking on the color picker causes the application to crash and display a blank white screen.

### To Reproduce

1. Open PDFme playground.
2. Select an element with a color property. (Eg. Text)
3. Click the color picker.
4. Observe the application.

### Expected behavior

The color picker should open normally and allow the user to select a color.

### Your Environment

```markdown
- pdfme package(@pdfme/generator or @pdfme/ui): 6.1.9
- pdfme version: 6.1.9
- Operating system: Mac
- Node.js version or Browser name & version: Chrome 149.0.7827.115
```

### Your Error Log

```shell
Uncaught TypeError: t.default.unstable_renderSubtreeIntoContainer is not a function
    at dist-DpxCzlU4.js:1385
```

### Additional context

The error occurs immediately after clicking the color picker.

It appears that a dependency is calling `ReactDOM.unstable_renderSubtreeIntoContainer`, which is no longer available in newer React versions. This may indicate a compatibility issue between the color picker component and the React version used by PDFme.

necessarylion:
<img width="1375" height="748" alt="Image" src="https://github.com/user-attachments/assets/adbc3eb9-1498-4601-93f6-73b7a3723716" />

odama626:
I just ran into this too.

react removed that api in react 19, so I think the color picker requires an older version of react than it's written for

hand-dot:
Thanks for the report. The React 19 color picker incompatibility was fixed in #1562.

We verified the published 6.1.12 packages with React 19: selecting a text field and opening its color picker works normally, without the removed ReactDOM API error or a blank screen.

Closing as fixed. Please upgrade the pdfme packages to 6.1.12 or later. If the problem persists, please open a new issue with the installed package versions and a minimal reproduction.

PR #1562, opened 2026-06-24
fix(ui): override form-render color widget for React 19 compat

necessarylion:
- #1561 

SCR-20260625-jwhx


hand-dot:
Thank you @necessarylion !

PR #1576, opened 2026-07-16
Fix performance: font deepcopy on renderKey

cubimon:
See https://github.com/pdfme/pdfme/issues/693

Fonts are copied on renderKey resulting in low performance if fonts are used as bytearrays instead of string urls.
This fix avoids copying the font data in renderKey.

### DIFF
--- a/packages/ui/src/components/Renderer.tsx
+++ b/packages/ui/src/components/Renderer.tsx
@@ -9,7 +9,7 @@ import {
   Plugin,
   UIOptions,
   cloneDeep,
-  Font,
+  type Font,
 } from '@pdfme/common';
 import { theme as antdTheme } from 'antd';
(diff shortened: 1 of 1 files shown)


raulopaulo (approved):

dharabhimani94 (approved):

dharabhimani94:
@hand-dot Hi! Could you please review this PR when you have a chance? 

I'm currently facing the same issue in our project, and it has become a significant performance bottleneck. From what I can see, this PR appears to address the problem, but I'd really appreciate your review to confirm that everything works as expected.

If the implementation looks good and passes your checks, could you please merge it? It would help us move forward and unblock our work. Thanks!

hand-dot:
on it

cubimon:
Thank you

hand-dot:
Released as 6.1.12.

PR #1577, opened 2026-07-22
fix(ui): only enter editing mode on click for text-editable schema types 

Karthi13krishna:
onClickMoveable unconditionally set editing=true on any click (no drag) on an already-selected element, which hides the Moveable control box (!editing && <Moveable ... />) with no replacement UI for non-text types. Clicking a selected image/shape/icon element a second time made it appear stuck — no resize/drag handles — until the selection was cleared and re-made from scratch.  Restrict setEditing(true) to selections where every active schema is text or multiVariableText, mirroring the existing schemaTypes check in the rotatable useMemo just above.


hand-dot:
Thank you for the PR. Clicking an already-selected image to enter edit mode and replace the image is intended behavior, so I’m closing this PR.

If there is a separate issue, such as being unable to interact with the image in edit mode, please share the steps to reproduce it. I’m happy to continue the discussion if needed.

PR #1597, opened 2026-09-05
fix(tables): omit undefined keys from mapCellStyle

### LINKED ISSUE #1573: mapCellStyle emits undefined-valued keys that overwrite built-in table style defaults
### Summary

When a Table schema's `headStyles`/`bodyStyles` object omits an optional style property (e.g. `padding`, `borderWidth`, `borderColor`), [`mapCellStyle`](https://github.com/pdfme/pdfme/blob/main/packages/schemas/src/tables/tableHelper.ts) emits that key with value `undefined`. Because the downstream merges use `Object.assign`/object spread, those `undefined` values **overwrite** the library's own defaults instead of falling through to them — producing `NaN` in cell geometry and broken/blank table rendering.

### Root cause

`mapCellStyle` (`packages/schemas/src/tables/tableHelper.ts`, ~L70–82) returns the mapped properties directly:

```ts
function mapCellStyle(style: CellStyle): Partial<Styles> {
  return {
    // ...
    lineWidth: style.borderWidth,   // undefined if the schema didn't set it
    cellPadding: style.padding,     // undefined if the schema didn't set it
  };
}
```

Its output flows into `cellStyles`, which merges it over the defaults:

```ts
const defaultStyle = { /* cellPadding: 5, lineWidth: 0, textColor: '#000000', ... */ };
return Object.assign(defaultStyle, otherStyles, rowStyles, colStyles);
```

`Object.assign` (and object spread) copy own-enumerable keys **including those whose value is `undefined`**. Since the `mapCellStyle` output is part of `otherStyles`, an unset `padding`/`borderWidth` ends up as `cellPadding: undefined` / `lineWidth: undefined`, which clobbers `defaultStyle.cellPadding = 5` / `lineWidth = 0`. The `undefined` then becomes `NaN` in the layout math.

### Why it isn't always seen

The UI designer always populates every style field, so it doesn't surface there. It reproduces when a Table schema is **built programmatically** (or migrated from an older template) with a partial style object.

### Reproduction

Generate a PDF from a `table` schema whose `headStyles`/`bodyStyles` omit `padding` and `borderWidth`. The table renders with collapsed/overlapping cells (NaN padding) instead of using the documented defaults (`cellPadding: 5`, `lineWidth: 0`).

### Expected behavior

Omitted style properties fall through to the built-in defaults.

### Proposed fix

Strip `undefined` values before returning from `mapCellStyle`, so unset properties fall through to the defaults:

```ts
function mapCellStyle(style: CellStyle): Partial<Styles> {
  const mapping = {
    fontName: style.fontName,
    alignment: style.alignment,
    verticalAlignment: style.verticalAlignment,
    fontSize: style.fontSize,
    lineHeight: style.lineHeight,
    characterSpacing: style.characterSpacing,
    backgroundColor: style.backgroundColor,
    textColor: style.fontColor,
    lineColor: style.borderColor,
    lineWidth: style.borderWidth,
    cellPadding: style.padding,
  };
  return Object.fromEntries(
    Object.entries(mapping).filter(([, value]) => value !== undefined),
  ) as Partial<Styles>;
}
```

### Environment

- `@pdfme/schemas` 6.1.11 (the same code is present on `main` at the time of filing)
- Reproduced via `@pdfme/generator`

We currently ship this fix as a local package patch; happy to open a PR if the approach looks right to the maintainers.

Steve0x2a:
When a Table schema's `headStyles` / `bodyStyles` omit optional fields such as `padding` or `borderWidth`, `mapCellStyle` still emitted those keys as `undefined`. Downstream `Object.assign` copied them over the library defaults (`cellPadding: 5`, `lineWidth: 0`), so `Cell.padding` threw and table geometry collapsed. The Designer always fills every style field, so this mainly hits programmatic or migrated schemas.

Drop undefined keys from `mapCellStyle` so omitted fields fall through to the defaults. The fallback `cellPadding` / `lineWidth` values are box dimensions so `Cell.padding` can read them.

Reported by @sbeisch.

Fixes #1573

### DIFF
--- a/packages/schemas/__tests__/tableHelper.test.ts
+++ b/packages/schemas/__tests__/tableHelper.test.ts
@@ -1,0 +1,131 @@
+import { getDefaultFont } from '@pdfme/common';
+import type { Schema } from '@pdfme/common';
+import { describe, expect, test } from 'vitest';
+import { createSingleTable } from '../src/tables/tableHelper.js';
+import type { TableSchema } from '../src/tables/types.js';
+
+const basePdf = {
+  width: 210,
+  height: 297,
+  padding: [10, 10, 10, 10] as [number, number, number, number],
+};
+
+const spacingSide = (value: unknown, side: 'top' | 'right' | 'bottom' | 'left'): unknown => {
+  if (typeof value === 'number') return value;
+  if (value && typeof value === 'object') {
+    return (value as Record<string, unknown>)[side];
+  }
+  return value;
+};
+
+const expectUniformSpacing = (value: unknown, expected: number) => {
+  expect(value).not.toBeUndefined();
+  expect(value).not.toBeNull();
+  for (const side of ['top', 'right', 'bottom', 'left'] as const) {
+    expect(spacingSide(value, side)).toBe(expected);
+  }
+};
+
+const expectFiniteGeometry = (table: Awaited<ReturnType<typeof createSingleTable>>) => {
+  expect(Number.isFinite(table.getHeight())).toBe(true);
+  expect(Number.isFinite(table.getWidth())).toBe(true);
+  for (const row of table.allRows()) {
+    expect(Number.isFinite(row.height)).toBe(true);
+    for (const column of table.columns) {
+      const cell = row.cells[column.index];
+      if (!cell) continue;
+      expect(Number.isFinite(cell.width)).toBe(true);
+      expect(Number.isFinite(cell.height)).toBe(true);
+      expect(Number.isFinite(cell.padding('top'))).toBe(true);
+      expect(Number.isFinite(cell.padding('right'))).toBe(true);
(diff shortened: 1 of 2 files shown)


hand-dot:
Thanks @Steve0x2a for the focused fix, and @sbeisch for the detailed report! Merged into main as d71e544.

After bringing the branch up to date, all GitHub Actions test jobs passed, including Playground: https://github.com/pdfme/pdfme/actions/runs/34803831000

The two regression tests in this PR passed, along with eight additional local review checks covering explicit undefined values, zero/empty values, input immutability, independent cell defaults, and multi-page tables with and without repeated headers. The two existing-style PDF samples we compared also rendered identically before and after the fix. The additional review checks were local only.

This resolves #1573. Thanks again for the contribution!

PR #1599, opened 2026-09-13
fix(common): prevent negative page index for overlapping shrinking tables

### LINKED ISSUE #1598: Dynamic table: placeUnitsOnPages indexes pages[-1] and throws when an element overlaps a shrinking table
### Version

`@pdfme/generator@6.1.11`, `@pdfme/common@6.1.11`, `@pdfme/schemas@6.1.11`, Node 22.

### What happens

`generate()` throws

```
TypeError: Cannot read properties of undefined (reading 'push')
    at placeUnitsOnPages (@pdfme/common/dist/index.js:1061:27)
    at processDynamicPage (@pdfme/common/dist/index.js:1091:18)
    at getDynamicTemplate (@pdfme/common/dist/index.js:1143:26)
```

whenever a page holds a **dynamic table whose rendered height is smaller than its designed
height**, and **another element's top edge lies inside the table's designed vertical span**.

### Minimal reproduction

No application code — this is the whole script:

```js
const { generate } = require('@pdfme/generator');
const { text, table } = require('@pdfme/schemas');
const box = (v) => ({ top: v, right: v, bottom: v, left: v });

const template = {
  basePdf: { width: 210, height: 297, padding: [15, 15, 15, 15] },
  schemas: [[
    // a dynamic table near the top, DESIGNED 52.9mm tall, rendering ~25mm with one row
    { name: 'tbl', type: 'table', position: { x: 15, y: 19.91 }, width: 150, height: 52.932,
      showHead: true, head: ['A', 'B', 'C'], headWidthPercentages: [30, 30, 40],
      content: JSON.stringify([['1', '2', '3']]),
      tableStyles: { borderColor: '#000000', borderWidth: 0.3 },
      headStyles: { fontName: 'Roboto', alignment: 'left', verticalAlignment: 'middle', fontSize: 13, lineHeight: 1, characterSpacing: 0, fontColor: '#ffffff', backgroundColor: '#2980ba', borderColor: '', borderWidth: box(0.1), padding: box(5) },
      bodyStyles: { fontName: 'Roboto', alignment: 'left', verticalAlignment: 'middle', fontSize: 13, lineHeight: 1, characterSpacing: 0, fontColor: '#000000', backgroundColor: '', borderColor: '#888888', borderWidth: box(0.1), padding: box(5), alternateBackgroundColor: '#f5f5f5' },
      columnStyles: {} },
    // an element whose top edge sits INSIDE the table's designed span
    { name: 'caption', type: 'text', content: 'CAPTION', position: { x: 15, y: 20 }, width: 70, height: 9,
      fontSize: 11, fontName: 'Roboto', alignment: 'left', verticalAlignment: 'top', lineHeight: 1.4,
      characterSpacing: 0, fontColor: '#000000', backgroundColor: '', rotate: 0, opacity: 1 },
  ]],
};

generate({ template, inputs: [{ tbl: JSON.stringify([['1', '2', '3']]), caption: 'CAPTION' }], plugins: { text, table } })
  .then((p) => console.log('OK', p.length))
  .catch((e) => console.log('THREW:', e.message, '\n  at', e.stack.split('\n')[1].trim()));
```

Expected: a one-page PDF. Actual: the `TypeError` above.

### Why

`getDynamicTemplate` sorts a page's elements by their designed top edge and shifts everything after
a table by `actual − designed`. That holds while the elements are stacked. When an element
**overlaps** the table, the shift can carry its Y below zero, and `placeUnitsOnPages` turns that Y
into a page index with a bare `Math.floor`:

```js
let currentPageIndex = Math.floor(startGlobalY / contentHeight);   // → -1
let currentYInPage   = startGlobalY % contentHeight;
if (currentYInPage < 0) currentYInPage = 0;                        // Y is clamped …
…
while (pages.length <= currentPageIndex) pages.push([]);           // no-op for -1
…
pages[currentPageIndex].push(newSchema);                           // pages[-1] is undefined
```

`currentYInPage` is already clamped two lines below; the page index is the half that was missed, and
the `while` guard cannot fire for a negative index.

### Suggested fix

```js
let currentPageIndex = Math.max(0, Math.floor(startGlobalY / contentHeight));
```

An over-shifted element then lands on the first page instead of taking the whole render down.

### Why it matters in practice

We reached this from a real user's saved template, not from a fuzzer. The **Designer allows free
placement**, so two elements can be overlapped in a single drag — and the result is a server-side
500 whose message names neither the template nor the element, which reads to the user as "export is
broken" and to the operator as an unattributable crash. We are working around it by clamping a
table's designed height to the gap above the next element, but the guard belongs upstream.

Happy to open a PR with the one-line change plus a regression test if that's welcome.

hand-dot:
## Summary
A shrinking dynamic table can shift an overlapping element above the content area and make `placeUnitsOnPages` access `pages[-1]`. Clamp the calculated page index to zero, consistent with the existing vertical-position clamp, so generation can finish.

Add layout regressions for shrinking to positive and zero heights, the clamped caption position, following-element spacing, and immutable template/input data. Add generator regressions that reopen PDFs and verify page counts for one and two inputs.

This preserves existing overlap layout rules; it does not promise collision-free placement.

Fixes #1598.

Thanks to @abdulrahim-tero for the reproduction and suggested guard, and @tevinch for the additional verification.

## Validation
- Updated the branch with main, including the Monaco playground fix from #1600.
- GitHub Actions at `4fefcc6f2a70f3efe4a0cda1af296632a44620a6`: all package, CLI, and playground checks passed.
- CodeQL and Cursor Bugbot passed with no inline findings.
- Manual review of the page-index guard and regression coverage found no blocking issues; `git diff --check` passed.
- The Claude review command was attempted but could not run because organization policy disables subscription access for Claude Code.

CI: https://github.com/pdfme/pdfme/actions/runs/34793282514


tevinch:
I reproduced the playground build failure on `ea0ca66`. The locked Monaco 0.56.0 package no longer resolves the old `monaco-editor/esm/vs/...` imports. Using its public entries gets the build past that error.

There is one related detail in `JsxPlayground`: `monaco.languages.typescript` is no longer populated, so the existing guard silently skips the JSX compiler options and extra declarations. Importing the TypeScript registration module directly preserves that setup.

In `CodeEditor.tsx`, replace the Monaco imports with:

```ts
import * as monaco from 'monaco-editor/editor';
import editorWorker from 'monaco-editor/editor/editor.worker?worker';
import 'monaco-editor/features/register.all';
import 'monaco-editor/languages/definitions/markdown/register';
import 'monaco-editor/languages/definitions/typescript/register';
import jsonWorker from 'monaco-editor/languages/features/json/json.worker?worker';
import 'monaco-editor/languages/features/json/register';
import 'monaco-editor/languages/features/typescript/register';
import tsWorker from 'monaco-editor/languages/features/typescript/ts.worker?worker';
```

Keep the existing worker mapping and `loader.config({ monaco })`. In `JsxPlayground.tsx`, add:

```ts
import * as typeScriptLanguage from 'monaco-editor/languages/features/typescript/register';
```

Then change `configureJsxEditor` to take no argument and remove its first two lines (`const typeScriptLanguage = ...` and the early-return guard). The compiler options and `addExtraLib` calls stay as they are.

With these two changes on macOS / Node 24.19.0, the workspace build, playground build, playground lint, `tsc --noEmit`, and all 44 playground tests passed. Lint still reports the existing warnings.

Scripted browser checks against a production-bundled fixture using the real components also covered JSON diagnostics and formatting, Markdown editing, read-only behavior, and the TypeScript worker. The JSX check verifies that the compiler setup actually ran, waits for newly edited two-page content to render, then generates a PDF; reopening it and extracting text confirmed the new text on both pages. These are local results; the upstream checks still need rerunning.

hand-dot:
Thanks for the detailed investigation and verification! These Monaco fixes, including feature registration and the JSX TypeScript setup, are now merged into main via #1600. The playground build and tests are passing, and I also verified completion suggestions and previews locally.

This PR still needs to be updated with main and have CI rerun.

